PR #1, opened 2017-11-24
The Basics

vext01:
Here is a small Rust library to trace any given pid with Intel PT.

It's just enough to get us going:
```
fn main() {
    let mut res = 0;
    let pid = unsafe { getpid() };

    let mut tracer = Tracer::new().trace_filename("simple_example.ptt");
    tracer.start_tracing();
    for i in 1..10 {
        res += i + pid;
    }
    tracer.stop_tracing();

    println!("result: {}", res);
}
```

 * Calls to C, storing the C state behind an opaque pointer.
 * For now writes trace to file (long term, to memory buffer).
 * If something goes wrong, panic (long term, `Result`s).
 * Doesn't make any attempt to pre-prime the tracer.
 * All public stuff documented.
 * Linted with Clippy.
 * Formatted with rustfmt.
 * Debugging statements left on for now.

Typical debug output:

```
$ sudo ./target/debug/examples/simple_example
src/traceme.c:345 [traceme_start_tracer]: target_pid=6401, trace_filename=simple_example.ptt, map_filename=simple_example.map, data_bufsize=64, aux_bufsize=1024
src/traceme.c:125 [stash_maps]: saving map to simple_example.map
src/traceme.c:385 [traceme_start_tracer]: Wait for tracer to init
src/traceme.c:250 [tracer_thread]: tracer init
src/traceme.c:314 [tracer_thread]: resume main thread
src/traceme.c:392 [traceme_start_tracer]: resume
src/traceme.c:404 [traceme_stop_tracer]: stopping tracer
src/traceme.c:417 [traceme_stop_tracer]: wait for trace thread to exit
src/traceme.c:214 [poll_loop]: wake
src/traceme.c:215 [poll_loop]: aux_head=  0x0000000960
src/traceme.c:216 [poll_loop]: aux_tail=  0x0000000000
src/traceme.c:217 [poll_loop]: aux_offset=0x0000266240
src/traceme.c:218 [poll_loop]: aux_size=  0x0004194304
src/traceme.c:173 [read_circular_buf]: read with no wrap
src/traceme.c:233 [poll_loop]: poll loop exit: awoke 1 times
src/traceme.c:320 [tracer_thread]: tracer thread exit
src/traceme.c:430 [traceme_stop_tracer]: tracing complete
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "traceme"
+version = "0.1.0"
+authors = ["Edd Barrett <vext01@gmail.com>"]
+
+[dependencies]
+libc = "0.2.33"
+
+[build-dependencies]
+gcc = "0.3.54"
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,45 @@
+// Copyright (c) 2017 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+extern crate gcc;
+
+fn main() {
+    gcc::Build::new()
+               .file("src/traceme.c")
+               .include("src")
+               .compile("traceme");
+}
--- a/examples/simple_example.rs
+++ b/examples/simple_example.rs
@@ -1,0 +1,59 @@
+// Copyright (c) 2017 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+extern crate traceme;
+extern crate libc;
+
+use traceme::Tracer;
+use libc::getpid;
+
+/// Trace a simple computation loop.
+///
+/// The result is printed to discourage the compiler from optimising the computation out.
+fn main() {
+    let mut res = 0;
+    let pid = unsafe { getpid() };
+
+    let mut tracer = Tracer::new().trace_filename("simple_example.ptt");
+    tracer.start_tracing();
+    for i in 1..10 {
+        res += i + pid;
+    }
+    tracer.stop_tracing();
+
+    println!("result: {}", res);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,182 @@
+// Copyright (c) 2017 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#![cfg_attr(feature="clippy", feature(plugin))]
+#![cfg_attr(feature="clippy", plugin(clippy))]
+
+extern crate libc;
+
+use libc::{pid_t, c_char, c_void, getpid, size_t};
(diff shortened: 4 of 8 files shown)


vext01:
My first Rust PR, so probably a fair amount of stuff to comment on.

Hope you like it though :)

ltratt:
Can I suggest you squash this a bit before review? It's much easier to review "proper" commits (so to speak).

vext01:
I can do, but it wont be any different to the "changed files" tab.

In other words this is all one commit.

ltratt:
Fair enough, but it's still a good discipline, so let's do it :)

vext01:
Splat!

(Future changes should be more incremental)

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+
+#define TRACE_OUTPUT    "trace.data"
+#define SYSFS_PT_TYPE   "/sys/bus/event_source/devices/intel_pt/type"
+#define MAX_PT_TYPE_STR 8
+
+#ifndef INFTIM
+#define INFTIM -1
+#endif
+
+/* Debug helpers */
```
vext01: Can probably kill this comment.

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+
+    DEBUG("tracer init");
+
+    /* Basic configuration */
+    memset(&attr, 0, sizeof(attr));
+    attr.size = sizeof(attr);
+    attr.size = sizeof(struct perf_event_attr);
+
+    /* Get the perf "type" for Intel PT */
+    pt_type_file = fopen(SYSFS_PT_TYPE, "r");
```
vext01: Missing close.

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+        err(EXIT_FAILURE, "sem_init");
+    }
+
+    /* Spawn a thread to deal with copying out of the PT aux buffer */
+    rc = pthread_create(&tr_ctx->tracer_thread, NULL, tracer_thread, (void *) tr_ctx);
+    if (rc) {
+        err(EXIT_FAILURE, "pthread_create");
+    }
+
+    DEBUG("Wait for tracer to init");
```
vext01: All the other debug prints start with a lower-case letter.

Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "traceme"
+version = "0.1.0"
+authors = ["Edd Barrett <vext01@gmail.com>"]
+
+[dependencies]
+libc = "0.2.33"
+clippy = {version = "*", optional = true}
```
ltratt: We don't want to add Clippy as a dependency. `cargo clippy` runs Clippy without the dependency.

Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[build-dependencies]
+gcc = "0.3.54"
+
+[features]
+default = []
```
ltratt: What does this do?

build.rs
```diff
@@ -0,0 +1,45 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+extern crate gcc;
+
+fn main() {
+    gcc::Build::new()
+        .file("src/traceme.c")
```
ltratt: This should look like:

```rust
gcc::Build::new()
           .file("...")
```

examples/simple_example.rs
```diff
@@ -0,0 +1,56 @@
...
+// SOFTWARE.
+
+extern crate traceme;
+extern crate libc;
+
+use traceme::Tracer;
+use libc::getpid;
+
+fn main() {
+    let mut res = 0;
```
ltratt: I don't understand what `res` is really doing. Is it just a random thing to prove a point or ...?

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+            data_bufsize: 64,
+            aux_bufsize: 1024,
+        }
+    }
+
+    /// Set where to write the trace.
+    ///
+    /// # Arguments
+    ///
+    /// * `filename` - Filename to use.
```
ltratt: I can't help feel this is not the most useful comment ever ;)

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn target_pid(mut self, pid: pid_t) -> Self {
+        self.target_pid = pid;
+        self
+    }
+
+    /// Set the PT data buffer size.
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - The size to use.
```
ltratt: Is there a suggested size or range of sizes? How do people know if they should set 8 or 10000000?

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn aux_bufsize(mut self, size: usize) -> Self {
+        self.aux_bufsize = size;
+        self
+    }
+
+    // Make the map filename by setting/adding a ".map" extension to the trace filename
+    fn make_map_filename(&self, trace_filename: &CString) -> CString {
```
ltratt: Is there an associated test for this method?

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        if !map_path.set_extension("map") {
+            panic!("failed to construct map filename");
+        }
+        let map_string = map_path.to_str().and_then(|ms| CString::new(ms).ok());
+        map_string.expect("failed to construct map filename")
+    }
+
+    /// Records execution of the selected PID into the chosen output file.
+    ///
+    /// Tracing continues until [stop_tracing](struct.Tracer.html#method.stop_tracing) is called.
```
ltratt: You can just say `until 'stop_tracing' is called` (where the `'` characters are reverse backticks) -- you don't need to do the full(ish) URL dance.

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn start_tracing(&mut self) {
+        if self.tracer_ctx.is_some() {
```
ltratt: It might be easier to just do:

```rust
assert!(self.tracer_ctx.is_none())
```

rather than this more lengthy `if` statement.

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+            trace_filename: self.trace_filename.as_ptr(),
+            map_filename: map_filename_c.as_ptr(),
+            data_bufsize: self.data_bufsize,
+            aux_bufsize: self.aux_bufsize,
+        };
+
+        // Call C
+        let conf_ptr = &tr_conf as *const TracerConf;
+        let opq_ptr = unsafe {
+            // By transmuting we avoid the need to make TracerConf public.
```
ltratt: Isn't there some way that doesn't involve transmuting?

src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+            traceme_start_tracer(transmute::<*const TracerConf, *const c_void>(conf_ptr))
+        };
+        self.tracer_ctx = Some(opq_ptr);
+    }
+
+    /// Turns off the tracer.
+    ///
+    /// [start_tracing](struct.Tracer.html#method.start_tracing) must have been called prior.
```
ltratt: Same URL comment as above.

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+
+    free(cmd);
+}
+
+/*
+ * Write part of a buffer to a file descriptor.
+ */
+static void
+write_buf_to_disk(int fd, void *buf, __u64 size) {
+    __u64 written = 0;
```
ltratt: I don't think this type is correct: I think it should be `ssize_t`.

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+    ssize_t res;
```
ltratt: Why is this variable defined here and not where it's first used below?

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+
+    /* Start disabled */
+    attr.disabled = 1;
+
+    /* No skid */
+    attr.precise_ip = 3;
+
+    /* Acquire file descriptor through which to talk to Intel PT */
+    tr_ctx->perf_fd = syscall(SYS_perf_event_open, &attr,
+        tr_ctx->target_pid, -1, -1, 0);
```
ltratt: Shouldn't this be indented to just after the open bracket?

src/traceme.c
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /*
+     * Allocate buffers.
+     *
+     * Data buffer is preceded by one management page, hence `1 + data_bufsize'.
+     */
+    page_size = getpagesize();
+    base = mmap(NULL, (1 + tr_ctx->data_bufsize) * page_size, PROT_WRITE,
+        MAP_SHARED, tr_ctx->perf_fd, 0);
```
ltratt: Same indentation comment as above.

ltratt:
Can we move the C code from the 1980s into the 1990s? There's no advantage to using such an old dialect of C: it's harder to read, and no compiler we'll ever encounter is so limited. Let's use C99 at least, use `//` comments, and put type definitions where they're first used, not all at the beginning of the function.

Cargo.toml
vext01: I was following the clippy instructions here:
https://github.com/rust-lang-nursery/rust-clippy#usage

`cargo clippy` only works if you have clippy installed, whereas the recommended way always works.

(Note that it's an optional dependency, so by default you won't get it.)

Cargo.toml
vext01: Again, taken from the clippy instructions.

Cargo.toml
vext01: Admittedly `cargo clippy` is easier, but I think we should leave these lines in for those that want to do it the other way.

build.rs
vext01: Are you sure? I followed the style from the gcc crate docs.
https://crates.io/crates/gcc

clippy doesn't complain about either of course :P

examples/simple_example.rs
vext01: It's the result of some computation that gets printed at the end. We might as well trace something that definitely can't be optimised out.

src/lib.rs
vext01: It depends how many events you want to collect and with what latency.

A small buffer will require more frequent copy outs, a large buffer will consume more memory. We are going to have to experiment with these when the time comes.

src/lib.rs
vext01: There isn't.

Do you really want me to test a function that turns `file.ptt` to `file.map` though? A bit extreme, no?

src/lib.rs
vext01: That's not right.

Now:


With:
```diff
-    /// Tracing continues until [stop_tracing](struct.Tracer.html#method.stop_tracing) is called.
+    /// Tracing continues until `stop_tracing` is called.
```



I'm certain you need the URL to have the clickable link, we spoke in #rust about this on Friday afternoon.

src/lib.rs
vext01: That's not quite the same though. We'd lose the reason why we failed.

I had intended to use `.expect`, but that isn't available for Booleans in Rust.

src/lib.rs
vext01: Heh, I knew you would flag this.

I spent a fair amount of time investigating this on Friday. It used to be the case that passing a pointer out to C would mean the type being pointed to would need to be public. This no longer seems to be the case after running rustup again, so I'll push a fix.

src/traceme.c
vext01: Because `write(2)` returns a `ssize_t`?

In general I agree, but in this case it's OK as we check the only case where we can have a negative result. By using a 64-bit unsigned we can count higher without an overflow (or in the case of `ssize_t`, undefined behaviour).

src/traceme.c
vext01: Nah, that's PEP8-style.

vext01:
The code style I used was "softdev KNF", which is what we agreed on for libkrun (https://github.com/softdevteam/krun/blob/30d9c6b4d57b497dc3eb13e02ddc72c911223ce2/libkrun/libkruntime.c#L48). Admittedly, I forgot to use `//` comments, but the requirement to define variables on their first use is new.

I don't mind changing the style, but you need to write it down somewhere and stick to it, because otherwise how should we know what style to use?

FWIW, I like to put my variables at the top when writing C. It was not because I expect to use an old compiler. I'm trying to cut down on that sort of thing :)

build.rs
vext01: I suppose your intuition is that you are calling `.file` on the result of what you line up with. I think I can buy that.

vext01:
Fixed my own comments, and yours which don't need a response.

Cargo.toml
ltratt: Let's not clog up the dependencies please (there's no advantage to doing so).

Cargo.toml
ltratt: OK, we can kill this then.

examples/simple_example.rs
ltratt: Maybe we can add a simple comment to make that clear?

build.rs
ltratt: Yep, correct.

src/lib.rs
ltratt: OK.

src/lib.rs
ltratt: Ironically I found this function a little difficult to reason about (given its small size and seeming simplicity).

src/lib.rs
ltratt: Oh... I stand corrected!

src/lib.rs
ltratt: You can add a reason to `assert`s (and it's really an `assert` IMHO).

src/traceme.c
ltratt: There's no guarantee that `ssize_t` fits in a `u64` AFAIK.

ltratt:
When you write Rust code, do you define all the `let`s at the top of the function? I bet not... because it makes code harder to read ;) So why do it in C? It made sense once up on a time, but we're waaaay beyond that now.

Cargo.toml
vext01: I disagree, but as you wish.

src/lib.rs
vext01: Yeah, Rust didn't make it easy.

src/lib.rs
vext01: Ah, I didn't know you could add a reason to assert, OK.

src/traceme.c
vext01: I'm not sure now, but given:

```
size_t is guaranteed to be at least 16 bits wide. Additionally, POSIX includes ssize_t, which is a signed integral type of the same width as size_t.
```
I think we can use `ssize_t` unless we find overflow problems later.

vext01:
Again, I disagree, but ok.

src/traceme.c
ltratt: 1024 bits is bigger than 16 bits, but won't fit in a `u64`...

vext01:
I think this is OK now? Can I squash?

vext01:
Try this then.

ltratt:
Yes, please squash away. Sorry for all the comments, but I think we've ended up with something quite nice :)

vext01:
splat

PR #5, opened 2017-11-28
Add Travis support.

vext01:
Travis test. Please ignore for now.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -36,10 +36,41 @@
 // SOFTWARE.
 
 extern crate gcc;
+use std::env;
+use std::process::Command;
+use std::collections::HashMap;
+
+fn run_lsb_release(args: Vec<&str>) -> HashMap<String, String> {
+    let output = Command::new("lsb_release")
+                         .args(args)
+                         .output()
+                         .unwrap();
+    assert!(output.status.success());
+    let output_s = String::from_utf8(output.stdout).unwrap();
+
+    let mut res = HashMap::new();
+    for line in output_s.lines() {
+        let parts: Vec<&str> = line.split(":").collect();
+        assert!(parts.len() == 2);
+        res.insert(String::from(parts[0].trim()), String::from(parts[1].trim()));
+    }
+    res
+}
 
 fn main() {
-    gcc::Build::new()
-               .file("src/traceme.c")
-               .include("src")
-               .compile("traceme");
+    let mut c_build = gcc::Build::new();
+    c_build.file("src/traceme.c").include("src");
+
+    // Travis' Linux setup is too old for what we do with linux-perf.
+    if env::var("TRAVIS").is_ok() {
+        // Crash out when Travis updates to a newer Ubuntu or a different OS.
+        // When this fires, travis may have a new enough linux-perf to test properly.
+        let lsb_rel = run_lsb_release(vec!["-ir"]);
+        assert!(lsb_rel["Distributor ID"] == "Ubuntu");
+        assert!(lsb_rel["Release"] == "14.04");
+        // Setting -DTRAVIS causes the C API to be stubbed.
+        c_build.define("TRAVIS", None);
+    }
+
+    c_build.compile("traceme");
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: rust
+rust: nightly
+cache: cargo
+
+before_script:
+    - export PATH="$PATH:$HOME/.cargo/bin"
+
+script:
+    - cargo build --examples
+    - cargo doc --no-deps
+    - cargo test
+    - ./target/debug/examples/simple_example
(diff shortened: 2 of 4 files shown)


vext01:
I don't know. I just copied and adapted the config from diffract.

We have more problems than that though. It looks like travis is too old for what we are doing with perf.

vext01:
This isn't going to work I'm afraid. The newest environment travis can offer is Ubuntu Trusty, which is pre PT support.

Here's the manual page for `perf_event_open` in Trusty:
http://manpages.ubuntu.com/manpages/trusty/en/man2/perf_event_open.2.html

Note the lack of aux buffer.

ltratt:
Does that mean we can't even get it to build?

vext01:
Even `cargo check` wants to build the C part, so yes, sorry.

vext01:
One hack would be to stub the C part if we are on travis.

ltratt:
Is stubbing the C out difficult? If so, maybe we just have to skip Travis for this repo. Maybe we should set bors up on a build machine of our own...

vext01:
We would basicallly put the whole contents of the C file inside an `#if !defined(TRAVIS)` and then try to convince cargo to pass `-D travis` to the C compiler if the `TRAVIS` environment is set.

Easy without Cargo, not sure about with.

I suggest I spend half an hour or so trying to figure this out before moving on. Plan?

ltratt:
No argument from me.

vext01:
Who likes to rock the party? :tada: 

I tried a few different things, this is the least foul of them. You OK with it?

Looking at the Travis logs, looks like travis doesn't run tests automatically.

vext01:
(lol, half an hour. More like 2 hours).

build.rs
```diff
@@ -36,10 +36,17 @@
...
-    gcc::Build::new()
-               .file("src/traceme.c")
-               .include("src")
-               .compile("traceme");
+    let mut c_build = gcc::Build::new();
+    c_build.file("src/traceme.c").include("src");
+
+    // Travis' Linux setup is too old for what we do with linux-perf.
+    // Setting -DTRAVIS causes the C API to be stubbed.
+    if env::var("TRAVIS").is_ok() {
```
ltratt: Should (can?) we check *which* version of Travis? If/when they upgrade, maybe we should have an `assert` (or similar) to alert us of this, and make us investigate building things fully?

vext01: We can check the output of `lsb_release -r` if we are on travis and panic if the version changed.

ltratt: Sounds like a plan!

vext01:
This seems to do the trick.

Passing on Travis. Here it is failing on bencher8 with `TRAVIS` in the env:
```
   Compiling traceme v0.1.0 (file:///home/vext01/research/traceme)
error: failed to run custom build command for `traceme v0.1.0 (file:///home/vext01/research/traceme)`
process didn't exit successfully: `/home/vext01/research/traceme/target/debug/build/traceme-0ac45483236dea7b/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'assertion failed: lsb_rel["Distributor ID"] == "Ubuntu"', build.rs:69:8
```

Swapping the order of the assertions to test the other one:
```
   Compiling traceme v0.1.0 (file:///home/vext01/research/traceme)
error: failed to run custom build command for `traceme v0.1.0 (file:///home/vext01/research/traceme)`
process didn't exit successfully: `/home/vext01/research/traceme/target/debug/build/traceme-0ac45483236dea7b/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'assertion failed: lsb_rel["Release"] == "14.04"', build.rs:69:8
```

build.rs
```diff
@@ -37,14 +37,38 @@
...
 use std::env;
+use std::process::Command;
+use std::collections::HashMap;
+
+fn run_lsb_release(args: Vec<&str>) -> HashMap<String, String> {
+    let output = Command::new("lsb_release")
+                         .args(args)
+                         .output()
+                         .expect("failed to run lsb_release");
+    assert!(output.status.success(), "lsb_release failed");
```
ltratt: Having one thing as an `expect` and one as an `assert` feels a bit odd? Maybe make both `expect`s and/-r unify the message? Or, to be honest, you could just `unwrap` them (I don't think putting `expect` everywhere is a good idea: it clogs up the source code).

vext01: If you don't think it's favourable to include why we crashed, then we can just unwrap. Yeah. I got into this habit as backtraces rarely work(ed?) on BSD.

ltratt: I think they might start working once the Rust port properly moves to LLVM and libunwind is fixed. Anyway, up to you.

vext01:
Is this what you meant?

You can't unwrap() a bool, so it has to either be an assert or some other conditional check.

ltratt:
I think so. If you're OK with it, please squash.

vext01:
splat.

PR #6, opened 2017-12-04
Error handling for the C code.

vext01:
This replaces all of the `err()` and `errx()` calls with error propagation.

It's quite a big change. Sorry about this. It's been a thing of nightmares to get this right. 

In short, I've minimised the amount of sharing between threads so that it becomes clear(er) by design who should be freeing/closing what.

I've tested by inserting artificial errors into the code at various places and checking the Rust assertions fire. I'm also going to try some static/dynamic analysers on this to see if we can tease out more bugs.

So while I am doing that, can you see anything obviously wrong with this?

Rust error handling will come as a separate PR (did you read the error handling section of the rust book btw?)

Cheers

### DIFF
--- a/analyse.mk
+++ b/analyse.mk
@@ -1,0 +1,15 @@
+# Analyse for errors
+
+BIN=target/debug/examples/simple_example
+# XXX work out a way to kill sudo
+VALGRIND=sudo valgrind --error-exitcode=1
+
+all: valgrind
+
+.PHONY: valgrind
+
+${BIN}:
+	cargo build --examples
+
+valgrind: ${BIN}
+	${VALGRIND} --track-fds=yes --leak-check=full ${BIN}
--- a/build.rs
+++ b/build.rs
@@ -70,6 +70,7 @@ fn main() {
         assert!(lsb_rel["Release"] == "14.04");
         // Setting -DTRAVIS causes the C API to be stubbed.
         c_build.define("TRAVIS", None);
+        println!("cargo:rustc-cfg=travis");
     }
 
     c_build.compile("traceme");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@
 
 extern crate libc;
 
-use libc::{pid_t, c_char, c_void, getpid, size_t};
+use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
 use std::ffi::CString;
 use std::path::PathBuf;
 
@@ -48,7 +48,7 @@ use std::path::PathBuf;
 #[link(name = "traceme")]
 extern "C" {
     fn traceme_start_tracer(conf: *const TracerConf) -> *const c_void;
-    fn traceme_stop_tracer(tr_ctx: *const c_void);
+    fn traceme_stop_tracer(tr_ctx: *const c_void) -> c_int;
 }
 
 // Struct used to communicate a tracing configuration to C. Must stay in sync with the C code.
@@ -170,6 +170,9 @@ impl Tracer {
         let opq_ptr = unsafe {
             traceme_start_tracer(conf_ptr as *const TracerConf)
         };
+        if cfg!(not(travis)) {
+            assert!(opq_ptr != std::ptr::null(), "traceme_start_tracer failed");
+        }
         self.tracer_ctx = Some(opq_ptr);
     }
 
@@ -177,8 +180,11 @@ impl Tracer {
     ///
     /// [start_tracing](struct.Tracer.html#method.start_tracing) must have been called prior.
     pub fn stop_tracing(&self) {
-        unsafe {
-            traceme_stop_tracer(self.tracer_ctx.expect("tracer wasn't started"));
+        let rc = unsafe {
+            traceme_stop_tracer(self.tracer_ctx.expect("tracer wasn't started"))
+        };
+        if cfg!(not(travis)) {
+            assert!(rc == 0, "traceme_stop_tracer failed");
         }
     }
 }
--- a/src/traceme.c
+++ b/src/traceme.c
@@ -53,6 +53,8 @@
 #include <linux/perf_event.h>
 #include <sys/ioctl.h>
 #include <semaphore.h>
+#include <assert.h>
+#include <stdbool.h>
 
 #define TRACE_OUTPUT    "trace.data"
 #define SYSFS_PT_TYPE   "/sys/bus/event_source/devices/intel_pt/type"
@@ -76,16 +78,10 @@
  */
 struct tracer_ctx {
     pthread_t           tracer_thread;      // Tracer thread handle.
-    sem_t               tracer_init_sem;    // Tracer init sync.
-    int                 loop_done_fds[2];   // Tells the poll loop to exit.
+    int                 stop_fd_wr;         // Close to halt trace thread.
+    int                 stop_fd_rd;         // Read end of stop_fd_wr pipe.
     int                 perf_fd;            // FD used to talk to the perf API.
-    pid_t               target_pid;         // PID to trace.
-    int                 out_fd;             // Trace written here.
-    struct perf_event_mmap_page
-                        *mmap_header;       // Aux buffer meta-data.
-    void                *aux;               // Aux buffer itself.
-    size_t              data_bufsize;       // Data buf size (in pages).
-    size_t              aux_bufsize;        // Aux buf size (in pages).
+    int                 out_fd;             // Trace packets written here.
 };
 
 /*
@@ -100,42 +96,66 @@ struct tracer_conf {
     size_t      aux_bufsize;        // Aux buf size (in pages).
 };
 
+/*
+ * Stuff used in the tracer thread
+ */
+struct tracer_thread_args {
+    int                 perf_fd;            // Perf notification fd.
+    int                 stop_fd_rd;         // Polled for "stop" event.
+    sem_t               *tracer_init_sem;   // Tracer init sync.
+    int                 out_fd;             // Trace packets written here.
+    size_t              data_bufsize;       // Data buf size (in pages).
+    size_t              aux_bufsize;        // Aux buf size (in pages).
+};
+
+
 // Private prototypes.
 #ifndef TRAVIS
-static void stash_maps(pid_t, const char *);
-static void write_buf_to_disk(int, void *, __u64);
-static void read_circular_buf(void *, __u64, __u64, __u64 *, int);
-static void poll_loop(struct tracer_ctx *);
+static bool stash_maps(pid_t, const char *);
+static bool write_buf_to_disk(int, void *, __u64);
+static bool read_circular_buf(void *, __u64, __u64, __u64 *, int);
+static bool poll_loop(int, int, int, struct perf_event_mmap_page *, void *);
 static void *tracer_thread(void *);
+static int open_perf(pid_t target_pid);
 #endif
 
 // Exposed Prototypes.
 struct tracer_ctx *traceme_start_tracer(struct tracer_conf *);
-void traceme_stop_tracer(struct tracer_ctx *tr_ctx);
+int traceme_stop_tracer(struct tracer_ctx *tr_ctx);
 
 
 #ifndef TRAVIS
 /*
  * Save linker relocation decisions so that you can later recover the
  * instruction stream from an on-disk binary.
+ *
+ * Returns true on success or false otherwise.
  */
-static void
+static bool
 stash_maps(pid_t pid, const char *map_filename)
 {
     DEBUG("saving map to %s", map_filename);
+    bool ret = true;
 
-    char *cmd;
+    char *cmd = NULL;
     int res = asprintf(&cmd, "cp /proc/%d/maps %s", pid, map_filename);
     if (res == -1) {
-        err(EXIT_FAILURE, "asprintf");
+        cmd = NULL; // cmd undefined after error.
+        ret = false;
+        goto clean;
     }
 
     res = system(cmd);
     if (res != 0) {
-        err(EXIT_FAILURE, "system");
+        ret = false;
+        goto clean;
     }
 
-    free(cmd);
+clean:
+    if (cmd) {
+        free(cmd);
+    }
+    return ret;
 }
 
 /*
@@ -143,11 +163,14 @@ stash_maps(pid_t pid, const char *map_filename)
  *
  * The implementation is a little convoluted due to friction between ssize_t
  * and __u64.
+ *
+ * Returns true on success or false otherwise.
  */
-static void
+static bool
 write_buf_to_disk(int fd, void *buf, __u64 size) {
     char *buf_c = (char *) buf;
     size_t block_size = SSIZE_MAX;
+
     while (size > 0) {
         if (block_size > size) {
             block_size = size;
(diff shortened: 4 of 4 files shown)


src/traceme.c
```diff
@@ -35,6 +35,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+
+// XXX don't share the whole tracer context with the tracer thread?
```
vext01: Kill this.

src/traceme.c
```diff
@@ -154,71 +174,84 @@ write_buf_to_disk(int fd, void *buf, __u64 size) {
...
     *tail_p = head;
+    return 0;
 }
 
 /*
  * Take trace data out of the AUX buffer.
  */
-static void
-poll_loop(struct tracer_ctx *tr_ctx)
+static int
```
vext01: Add error description in comment.

src/traceme.c
```diff
@@ -273,49 +313,104 @@ tracer_thread(void *arg)
...
+ * Set up Intel PT buffers and start a poll() loop for reading out the trace.
+ */
+static void *
+tracer_thread(void *arg)
+{
+    DEBUG("tracer init");
+
+    /* Copy args in */
+    struct tracer_thread_args thr_args;
+    memcpy(&thr_args, (struct tracer_thread_args *) arg, sizeof(thr_args));
```
vext01: We are copying a pointer to a string on the rust stack here. That pointer becomes invalid after `start_tracing` in rust has returned. Although use of the pointer is synchronised, and we don't use the pointer after, I wonder if we should try and eliminate that pointer hanging around...

vext01: The pointer is the output filename, btw. We could open the file in the caller, and pass a fd instead. I'd be inclined to have the parent close the fd too if that were our plan.

vext01:
The merge conflict is easy to fix. I'll do this last.

vext01:
OK, looks like we might need a compile-time feature to distinguish an error from the C part being stubbed.

I'll continue with this tomorrow.

vext01:
This is ready for review. Checked with valgrind for fd and heap leaks. All OK.

ltratt:
Can you squash this down a bit before I try reviewing please?

vext01:
splat.

vext01:
I've just figured out how to allow `mmap`ing of the perf fd as a normal user too. We can probably kill the sudo thing now, and add a comment near the mmap that would fail.

analyse.mk
```diff
@@ -0,0 +1,15 @@
+# Analyse for errors
+
+BIN=target/debug/examples/simple_example
+# XXX work out a way to kill sudo
+VALGRIND=sudo valgrind --error-exitcode=1
+
+all: valgrind
+
+.PHONY: ${BIN} valgrind-memcheck valgrind-fd-check
```
vext01: Bah, just noticed this line is incorrect too. It should be:
```
.PHONY: valgrind
```

vext01:
Let's do the "no root" changes in a separate PR. This one is too big already.

build.rs
```diff
@@ -70,6 +70,8 @@ fn main() {
         assert!(lsb_rel["Release"] == "14.04");
         // Setting -DTRAVIS causes the C API to be stubbed.
         c_build.define("TRAVIS", None);
+        // Set a Rust compile-time switch too.
```
ltratt: This comment isn't very useful (what does the switch do?). Could be removed?

src/lib.rs
```diff
@@ -170,15 +170,23 @@ impl Tracer {
         let opq_ptr = unsafe {
             traceme_start_tracer(conf_ptr as *const TracerConf)
         };
+        if cfg!(not(travis)) {
+            // Under Travis the whole C library is stubbed and returns a NULL ptr.
```
ltratt: This comment feels like it needs to be inverted (at first I read it as "this assert is only valid on Travis")?

src/lib.rs
```diff
@@ -170,15 +170,23 @@ impl Tracer {
...
     ///
     /// [start_tracing](struct.Tracer.html#method.start_tracing) must have been called prior.
     pub fn stop_tracing(&self) {
-        unsafe {
-            traceme_stop_tracer(self.tracer_ctx.expect("tracer wasn't started"));
+        let rc = unsafe {
+            traceme_stop_tracer(self.tracer_ctx.expect("tracer wasn't started"))
+        };
+        if cfg!(not(travis)) {
+            // Under Travis stub returns -1.
```
ltratt: Same as above.

src/traceme.c
```diff
@@ -158,71 +180,86 @@ write_buf_to_disk(int fd, void *buf, __u64 size) {
...
-            DEBUG("aux_size=  0x%010llu", mmap_header->aux_size);
-
-            read_circular_buf(tr_ctx->aux, mmap_header->aux_size,
-                head, &mmap_header->aux_tail, tr_ctx->out_fd);
+            DEBUG("aux_tail=  0x%010llu", mmap_hdr->aux_tail);
+            DEBUG("aux_offset=0x%010llu", mmap_hdr->aux_offset);
+            DEBUG("aux_size=  0x%010llu", mmap_hdr->aux_size);
+
+            if (read_circular_buf(aux, mmap_hdr->aux_size,
+                head, &mmap_hdr->aux_tail, out_fd) == -1) {
```
ltratt: The formatting of this call makes it very hard to read w.r.t. the next line.

src/traceme.c
```diff
@@ -277,49 +321,103 @@ tracer_thread(void *arg)
...
+
     /*
      * Allocate buffers.
      *
      * Data buffer is preceded by one management page, hence `1 + data_bufsize'.
      */
-    int page_size = getpagesize();
-    void *base = mmap(NULL, (1 + tr_ctx->data_bufsize) * page_size, PROT_WRITE,
-        MAP_SHARED, tr_ctx->perf_fd, 0);
+    base = mmap(NULL, base_size, PROT_WRITE,
```
ltratt: Not sure why this call is split on two lines?

src/traceme.c
```diff
@@ -277,49 +321,103 @@ tracer_thread(void *arg)
...
+        MAP_SHARED, thr_args->perf_fd, header->aux_offset);
     if (aux == MAP_FAILED) {
-        err(EXIT_FAILURE, "mmap2");
+        ret = -1;
+        goto clean;
     }
-    tr_ctx->aux = aux;
 
-    // Resume the interpreter loop.
+    // Copy out arguments for the poll loop.
```
ltratt: I assume this comment should really be something like "copy out because soon we'll lose control over the struct and the values will be overwritten"?

src/traceme.c
```diff
@@ -277,49 +321,103 @@ tracer_thread(void *arg)
...
+    if (base) {
+        if (munmap(base, base_size) == -1) {
+            ret = -1;
+        }
+    }
+    if (!sem_posted) {
+        assert(ret == -1);
+        sem_post(thr_args->tracer_init_sem);
+    }
+
```
ltratt: Extranenous blank line.

src/traceme.c
```diff
@@ -277,49 +321,103 @@ tracer_thread(void *arg)
...
+        ret = -1;
+        goto clean;
+    }
 
-    DEBUG("tracer thread exit");
-    pthread_exit(NULL);
+clean:
+    if (aux) {
+        if (munmap(aux, header->aux_size) == -1) {
+            ret = -1;
```
ltratt: What's the point of setting `ret` at this point?

src/traceme.c
```diff
@@ -347,54 +447,128 @@ traceme_start_tracer(struct tracer_conf *tr_conf)
     DEBUG("Travis mode. Not starting tracing");
     return NULL;
 #else
+    int failing = 0;
+    struct tracer_ctx *tr_ctx = NULL;
+    int perf_fd = -1, out_fd = -1;
+    sem_t tracer_init_sem;
```
ltratt: Variable defined before use.

src/traceme.c
```diff
@@ -347,54 +447,128 @@ traceme_start_tracer(struct tracer_conf *tr_conf)
     DEBUG("Travis mode. Not starting tracing");
     return NULL;
 #else
+    int failing = 0;
```
ltratt: This should be a `bool` I think? [we are using C99 :)]

src/traceme.c
```diff
@@ -403,39 +577,57 @@ traceme_start_tracer(struct tracer_conf *tr_conf)
...
 
     // Wait for poll loop to exit.
     DEBUG("wait for trace thread to exit");
-    if (pthread_join(tr_ctx->tracer_thread, NULL) != 0) {
-        err(EXIT_FAILURE, "pthread_join");
+    void *thr_exit;
+    if (pthread_join(tr_ctx->tracer_thread, &thr_exit) != 0) {
+        ret = -1;
+    }
+    if ((intptr_t) thr_exit != 0) {
```
ltratt: I was sort-of expecting `thr_exit` to be dereferenced, because aren't you just testing whether the pointer you've received is `NULL` or not (which it will never be due to the nature of your calls to `pthread_exit`).

build.rs
vext01: It makes `!cfg(travis)` true. We can kill it if you want.

ltratt: I think either remove it or explain what it does a bit more precisely (I don't mind which).

src/lib.rs
vext01: It's the right way around. On travis we expect a NULL ptr. We coulsd add an else if you like.

src/traceme.c
vext01: This is how it's done in KNF. How should I format it?

src/traceme.c
vext01: Yes, that's explained in the next comment down.

ltratt: It's a bit late by that point :)

src/traceme.c
ltratt: I suggest we use the Rust idea: `head` would be aligned with `aux`.

src/traceme.c
vext01: Because this code is executed if we succeed and we want failure to clean up to be an error if you ask me.

src/traceme.c
vext01: yup

src/lib.rs
ltratt: What I'm saying is: the comment is confusing, because the comment refers to the implict `else` block. If we say something like "We can only check for XXX when not on Travis", it might be clearer?

src/traceme.c
ltratt: But we don't use `ret` past this point AFAICS.

src/traceme.c
vext01: Can do. Do you want all the functions that return an int for an error a bool too?

src/traceme.c
vext01: OK. FWIW, I still find this style quite jarring for C code.

src/traceme.c
vext01: `pthread_exit` only allows you to return a void pointer:
```
     void
     pthread_exit(void *value_ptr);
```
but we want to return an int. By using a `intptr_t` we can safely cast our void pointer to the int. It's not really a pointer.

build.rs
vext01: It's standard rust stuff, so I'll kill it.

src/lib.rs
vext01: I find that confusing. Shall we just kill the comment?

src/traceme.c
vext01: OK, so we really do need to draft up a C code style document.

vext01: It would also be nice if we can find a vim config to do that alignment automatically.

src/traceme.c
ltratt: Yes, I got that -- but I think you need to do `if (*((intptr_t *) thr_exit) != 0)` or similar.

src/traceme.c
ltratt: Good luck: I can't even get vim to reliably indent normal code properly!

src/lib.rs
ltratt: As you prefer.

src/traceme.c
vext01: We use it to decide what to return later.

https://github.com/softdevteam/traceme/blob/9b62089022573b92fbf4d1b4aac811c2f5b390e4/src/traceme.c#L414-L420

There was a reason I didn't return `(intptr_t *) ret`. Let me check.

vext01: Ah yes:
```
warning: src/traceme.c:422:18: warning: cast to 'intptr_t *' (aka 'long *') from smaller integer type 'int' [-Wint-to-pointer-cast]
warning:     pthread_exit((intptr_t *) ret);
warning:                  ^
warning: 1 warning generated.
```

ltratt: Don't you want `pthread_exit((intptr_t *) &ret)`?

src/traceme.c
vext01: it's `:set cino+=(0`

src/traceme.c
vext01: Eh? That would make a dangling pointer. Remember, we are using the pointer value as an integer.

vext01: We want to return `0` for success or `-1` for failure. We just happen to pack it into a void pointer as that's all the pthread interface gives us.

`&ret` would always indicate non-zero, as the address of ret won't ever be null.

ltratt: Ah, I see what's going on: it's the "`void *` is guaranteed to be able to hold any other value" trick. A bit evil, but kind of makes sense.

vext01: It's either that or use a real pointer and find somewhere for what it points to to live.

It's a shortcoming of the pthread api due to C's type system really.

ltratt: Right, but the problem then becomes that we're *also* abusing `intptr_t` and pretending that 0 and 1 are addresses. Why don't we just be honest and cast an `int` to a `void *` somehow (avoiding the compiler warning)?

vext01: I'm not sure there's a portable way. I mean, I could cast `int` to `long long` and then to `void *`, but this assumes x86_64.

As far as I can see, `intptr_t` is exactly what we want. It was introduced into C99 as a:
> [signed] integer type capable of holding a pointer

http://en.cppreference.com/w/c/types/integer

src/traceme.c
vext01: So I think we solved this above.

src/traceme.c
vext01: I think we should not use bools at the API level, as this places an unnecessary constraint on the consuming C code. I've used a bool inside as you suggested though.

vext01:
So if you are happy with `intptr_t`, I think that's all your comments addressed.

src/traceme.c
ltratt: We're never going to come across a compiler that doesn't at least understand C99 (an 18 year old standard remember!), so let's please use bools when they're the right datatype.

src/traceme.c
ltratt: Another dumb question: shouldn't we be using `intptr_t` and not `intptr_t *`?

src/traceme.c
vext01: so return false on error?

src/traceme.c
vext01: oh. good question.

src/traceme.c
ltratt: I think it would be clearer that way.

vext01: Hrm.

It looks like the libc crate doesn't have these. I was going to raise a bug, but it looks like the rust devs are well aware. 

https://github.com/rust-lang/rust/issues/14608
https://github.com/rust-lang/rfcs/issues/992

We can probably use one of the hacks show here:
https://github.com/rust-lang/rfcs/issues/992#issuecomment-264631983

But we need to be careful, because what these guys are implying is that the integer values under `true` and `false` are platform dependent! I didn't see that coming.

src/traceme.c
vext01: I don't think we are allowed to cast 0 and 1 to intptr_t:
```
warning: src/traceme.c:418:22: warning: expression which evaluates to zero treated as a null pointer constant of type 'void *' [-Wnon-literal-null-conversion]
warning:         pthread_exit((intptr_t) 0);
warning:                      ^~~~~~~~~~~~
warning: src/traceme.c:421:22: warning: incompatible integer to pointer conversion passing 'intptr_t' (aka 'long') to parameter of type 'void *' [-Wint-conversion]
warning:         pthread_exit((intptr_t) -1);
warning:                      ^~~~~~~~~~~~~
warning: /usr/include/pthread.h:242:33: note: passing argument to parameter '__retval' here
warning: extern void pthread_exit (void *__retval) __attribute__ ((__noreturn__));
warning:                                 ^
warning: 2 warnings generated.
```

Any other ideas? All the examples on the net cast to `void *`, which seems to work.

Let me think about this a bit more...

src/traceme.c
ltratt: I doubt there's any platform where they're not 1 and 0 respectively, but typical C not to actually define that :/

vext01: OK, let's define them exactly as that and not rely on the rust bool being the same by chance.

src/traceme.c
vext01: I think we can actually kill the explicit calls to `pthread_exit`.  Negative numbers should be fine as long as you cast to the same type on the other side, but if you wanted bools, we could use a bool here too.

ltratt: I'm nervous about negative numbers because funny things might happen due to our not fully understanding the various casting rules. Best not to risk it IMHO.

If we don't need the calls, then let's kill them. If we need them, maybe we should just define constants like `THREAD_SUCCESS` / `THREAD_FAILURE` and so on? Then it doesn't matter what (positive) integer we set them to.

vext01: You are over-thinking it :) stdbool is basically an unsigned int.

vext01:
Try this.

src/traceme.c
ltratt: Yes, but what if we want other return codes? Or will success/failure only ever be it?

src/traceme.c
```diff
@@ -335,7 +335,7 @@ open_perf(pid_t target_pid) {
 /*
  * Set up Intel PT buffers and start a poll() loop for reading out the trace.
  *
- * Thread exits with 0 on success and -1 on error.
+ * Returns true or false otherwise.
```
vext01: Ugh. Sorry. Fix coming.

src/traceme.c
```diff
@@ -375,12 +375,15 @@ tracer_thread(void *arg)
         goto clean;
     }
 
-    // Copy out arguments for the poll loop.
+    /*
+     * Copy arguments for the poll loop, as when we resume the parent thread,
```
ltratt: This should be `//` commented (so it'll become 2 lines instead of 4).

ltratt:
OK, one minor comment, then I think we're ready to go.

src/traceme.c
vext01: Eh? So we can't use multi-line comments?

ltratt: Two lines of `//` is much easier to read than 4 lines of `/*...*/`.

vext01: I'm confused. We have multi-line comments all over. What is it about this one that is special?

ltratt: It's not about allowing multiline comments. It's about freeing ourselves from some of pre-C99 ugliness. Which is easier to read overall:

```C
/*
 * blah
 * blah
*/
```

or:

```C
// blah
// blah
```

Vertical space is at a premium, especially with the move to widescreen monitors, so the latter is preferable. I can just about tolerate the former use case before function definitions, even though it's a bit archaic, but within functions, let's just use `//`. [Another way of asking this is: would you use `/*...*/` for a two line comment in Rust? I reckon not.]

As to why this one: I noticed it. I'd happily never see `/* ... */` in a PR ever again ;)

vext01: I don't mind either.

If you really hate `/* */` though, then let's kill them all. At least then I know what style to use.

vext01:
I *think* this is the right commenting style now.

ltratt:
Looks good to me! If you squash, I can merge.

vext01:
Splat.

ltratt:
Minor typo on that second commit "too" -> "to".

vext01:
fixed.

PR #9, opened 2017-12-07
Use a bool return type for traceme_stop_tracer().

vext01:
This is my attempt to use a bool for the return type of one C function we missed.

I've tried to be careful about the conversion over the Rust/C boundary. The problem is, `bool` is of undefined size, meaning we don't know what to use to represent it on the Rust side.

I *thought* that by using a `u8` we are safe, but then I saw this comment:
https://github.com/rust-lang/rfcs/issues/1982#issuecomment-297531453

If a bool is 1-bit, then using a `u8` will read memory we don't own, argh. That said, this change does seem to work.

We don't have to merge this. We could use an `int` to be safe and the semantics are clear.

Any thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,15 +40,22 @@
 
 extern crate libc;
 
-use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
+use libc::{pid_t, c_char, c_void, getpid, size_t};
 use std::ffi::CString;
 use std::path::PathBuf;
+
+#[allow(dead_code)]
+#[repr(u8)]
+enum CBool {
+    CFalse = 0,
+    CTrue = 1,
+}
 
 // FFI stubs
 #[link(name = "traceme")]
 extern "C" {
     fn traceme_start_tracer(conf: *const TracerConf) -> *const c_void;
-    fn traceme_stop_tracer(tr_ctx: *const c_void) -> c_int;
+    fn traceme_stop_tracer(tr_ctx: *const c_void) -> u8;
 }
 
 // Struct used to communicate a tracing configuration to C. Must stay in sync with the C code.
@@ -184,7 +191,7 @@ impl Tracer {
             traceme_stop_tracer(self.tracer_ctx.expect("tracer wasn't started"))
         };
         if cfg!(not(travis)) {
-            assert!(rc == 0, "traceme_stop_tracer failed");
+            assert!(rc == CBool::CTrue as u8, "traceme_stop_tracer failed");
         }
     }
 }
--- a/src/traceme.c
+++ b/src/traceme.c
@@ -121,7 +121,7 @@ static int open_perf(pid_t target_pid);
 
 // Exposed Prototypes.
 struct tracer_ctx *traceme_start_tracer(struct tracer_conf *);
-int traceme_stop_tracer(struct tracer_ctx *tr_ctx);
+bool traceme_stop_tracer(struct tracer_ctx *tr_ctx);
 
 
 #ifndef TRAVIS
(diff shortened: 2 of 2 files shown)


vext01:
I actually don't understand how you could read 1-bit without reading at least a byte first.

src/lib.rs
```diff
@@ -40,15 +40,22 @@
 
 extern crate libc;
 
-use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
+use libc::{pid_t, c_char, c_void, getpid, size_t};
 use std::ffi::CString;
 use std::path::PathBuf;
 
+#[allow(dead_code)]
```
ltratt: Why do we need this? Isn't it also a module level directive, not just attached to an `enum`?

vext01: it's because we don't use `CFalse`, but we kinda do want it's definition.

```
warning: variant is never constructed: `CFalse`
  --> src/lib.rs:50:5
   |
50 |     CFalse = 0,
   |     ^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default
```

ltratt: Is the directive per-enum or per-module?

vext01: it's just for this enum. If I had put it right at the top of the file and prepended with a `!` it would have been for the whole module.

```
#[allow(dead_code)]    
#[repr(u8)]    
enum CBool {    
    CFalse = 0,    
    CTrue = 1,    
}    
    
enum J {    
    X = 1,    
}    
```

```
vext01@bencher8:~/research/traceme$ cargo build --examples
   Compiling traceme v0.1.0 (file:///home/vext01/research/traceme)
warning: enum is never used: `J`
  --> src/lib.rs:54:1
   |
54 | enum J {
   | ^^^^^^
   |
   = note: #[warn(dead_code)] on by default
```

ltratt: OK, I did not know one could do that.

ltratt:
I think this is a merge as-is job?

vext01:
Can you comment on the use of `u8` to represent a C `bool` on the Rust side? I'm not sure if it's correct.

src/lib.rs
```diff
@@ -40,15 +40,22 @@
...
+    CFalse = 0,
+    CTrue = 1,
+}
+
 // FFI stubs
 #[link(name = "traceme")]
 extern "C" {
     fn traceme_start_tracer(conf: *const TracerConf) -> *const c_void;
-    fn traceme_stop_tracer(tr_ctx: *const c_void) -> c_int;
+    fn traceme_stop_tracer(tr_ctx: *const c_void) -> u8;
```
ltratt: I guess this is `u8` because the Rust FFI doesn't support bools? If so, are we sure that bools will only ever be 0 or 1? If so, I think this is fine.

vext01: The Rust FFI doesn't support C bools because Rust doesn't know how big a `bool` is on any given platform. I'm hoping that if a `bool` were really (e.g.) a 4-byte `int` then our use of `u8` would cause it to be truncated to the least significant byte...

Does that make sense?

vext01: On the other hand, on the post linked above people are saying a C `bool`, could be 1-bit, so we'd be hoping that the Rust FFI does the necessary zero extension :\

I mean, the code works, but I'm not certain it's correct.

ltratt: Why don't we add a check (into a `build.rs`?) that compiles a couple of test cases for this in C so that we have some confidence that it'll keep working into the future (and, if it stops working, we'll be alerted)?

ltratt: OK, `build.rs` would be a bad idea, as they'd be checked all the time. I wonder if there's a way to do a sort of `configure.rs` equivalent?

vext01: You can do whatever you want in `build.rs` and then set `!cfg` macros, so you can somewhat emulate a configure script. I'd rather avoid anything like that if we can though.

FWIW, `sizeof(bool)` is 1 on my desktop and bencher8.

Here are my two concrete suggestions at this point:
 * Roll with the diff we have, because it will explode loudly at some point if we go it wrong.
 * Or, just use an `int` until rustc gets a well-defined `libc::c_bool`

I'll take either :)

ltratt: My point is more that `build.rs` is invoked regularly, so I'm not sure we would want to do invoke two C compiles every time we do `cargo check`. Although maybe we can cache results somewhere?

Unfortunately, I do think we should add a check that our assumptions are right :(

vext01:
Morning,

Let's wait a bit on this one. I have a SO question up, and I'm about 90%
certain u8 is what we want. I'm also going to ask aroind in IRC.

In the meantime, the Rust errors branch can be done independently.

vext01:
Well, I didn't get a conclusive answer sadly.

Can you take a look at this discussion and have a think:
https://stackoverflow.com/questions/47705093/what-is-the-correct-type-for-returning-a-c99-bool-to-rust-via-the-ffi

FWIW, my preferences, in order of most to least preferred:
 * Leave it an int until Rust is ready.
 * Use a Rust `bool` (which is now, and is likely to be standardised to be, a C99 `_Bool`)
 * Use feature checks and macros to alias an appropriate integer type to our own `c_bool`.

I like the first option because it's both fool and future proof.

ltratt:
When you say "leave it an int", what do you mean? The Rust or C side or ...?

vext01:
Leave the C signature of this function as it is, returning a C int, as this is well defined on both sides.

ltratt:
Regrettably I think you're right :( OK, let's do it.

vext01:
OK, so this PR is void for now then. Shall we close it?

ltratt:
OK.

vext01:
(We can also leave it open in the hope that `bool` is standardised soon)

vext01:
(up to you)

ltratt:
Should we keep this open or ...?

ltratt:
@vext01 Keep or kill?

vext01:
This is blocked until Rust makes a decision about the ABI of a bool.

It's also already bit-rotted, and not a hard change. Let's close this.

PR #11, opened 2017-12-08
Error handling on the Rust side.

vext01:
Here's the error handling for the Rust side.

I follow up PR will simplify some of the error handling by re-structuring slightly the tracer context.

OK?

### DIFF
--- a/examples/simple_example.rs
+++ b/examples/simple_example.rs
@@ -48,12 +48,17 @@ fn main() {
     let mut res = 0;
     let pid = unsafe { getpid() };
 
-    let mut tracer = Tracer::new().trace_filename("simple_example.ptt");
-    tracer.start_tracing();
+    let mut tracer = Tracer::new()
+                            .trace_filename("simple_example.ptt")
+                            .unwrap();
+    tracer.start_tracing().unwrap_or_else(|e| {
+        panic!(format!("Failed to start tracer: {}", e));
+    });
+
     for i in 1..10 {
         res += i + pid;
     }
-    tracer.stop_tracing();
 
+    tracer.stop_tracing().unwrap();
     println!("result: {}", res);
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,9 +40,12 @@
 
 extern crate libc;
 
+mod errors;
+
 use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
 use std::ffi::CString;
 use std::path::PathBuf;
+use errors::TraceMeError;
 
 // FFI stubs
 #[link(name = "traceme")]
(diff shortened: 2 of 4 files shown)


vext01:
I've just realised that the error types not containing anything don't need to be tuples. Commit coming.

examples/simple_example.rs
```diff
@@ -48,12 +48,19 @@ fn main() {
     let mut res = 0;
     let pid = unsafe { getpid() };
 
-    let mut tracer = Tracer::new().trace_filename("simple_example.ptt");
-    tracer.start_tracing();
+    let mut tracer = Tracer::new()
+                            .trace_filename("simple_example.ptt")
+                            .unwrap_or_else(|e| panic!(format!("{}", e)));
```
ltratt: What's the difference between this line and just saying `unwrap()`?

src/lib.rs
```diff
@@ -101,9 +104,9 @@ impl Tracer {
     /// # Arguments
     ///
     /// * `filename` - The filename in which to store trace packets.
-    pub fn trace_filename(mut self, filename: &str) -> Self {
-        self.trace_filename = CString::new(filename).expect("bad trace filename");
-        self
+    pub fn trace_filename(mut self, filename: &str) -> Result<Self, TraceMeError> {
```
ltratt: Is there a particular reason why `trace_filename` returns `Self`? Is it part of a "fluent" interface or ...?

ltratt:
Although I think it's a bad idea, unfortunately `try!` has been superseded by `?`, so we should use that in new code. [I think the additional syntax was not a great idea, but such is life.]

examples/simple_example.rs
vext01: Ah, does `unwrap()` print the error message then? If so, we can just unwrap.

vext01:
Ah. I'd forgotten to use `?`. It was introduced after v1 of the book was written. I can live with it, and may come to prefer it.

examples/simple_example.rs
vext01: It does. I pushed a change to simply unwrap.

vext01:
Any further comments?

src/lib.rs
vext01: I'm not sure what a fluent interface is, but I was following the rust paradigm where you construct and configure an object in a chain:

```
let mything = Thing().setting1("blah").setting2("flibble");
```

My next PR will simplify this particular piece of code, so there is not possibility of failure in the chain.

vext01: Looks like I just described a fluent interface :) Yes, a fluent interface, that's what I meant.

ltratt: OK. I just wanted to check you did it deliberately :)

ltratt:
I think this is ready for squashing.

vext01:
Like so.

ltratt:
Just to be clear: this is squashed and ready for merging?

vext01:
Correct. One commit is an unrelated fix, thus two commits.

PR #13, opened 2017-12-08
Simplify trace_filename error handling and ensure it has a `.ptt' extension.

vext01:
By using a `String` until the last moment when we hand off to C, we can avoid early `NulErrors` and also simplify `make_map_filename` quite a bit.

While we are here, check the trace filename ends in `.ptt`, and add a very basic test.

Makes sense?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,7 @@ struct TracerConf {
 #[derive(Default)]
 pub struct Tracer {
     /// Filename to store the trace to.
-    trace_filename: CString,
+    trace_filename: String,
     /// PID to trace.
     target_pid: pid_t,
     /// Data buffer size, in pages. Must be a power of 2.
@@ -87,11 +87,11 @@ impl Tracer {
     /// ```
     /// use traceme::Tracer;
     ///
-    /// Tracer::new().trace_filename("mytrace.ptt").unwrap().data_bufsize(1024).target_pid(666);
+    /// Tracer::new().trace_filename("mytrace.ptt").data_bufsize(1024).target_pid(666);
     /// ```
     pub fn new() -> Self {
         Tracer {
-            trace_filename: CString::new("traceme.ptt").unwrap(),
+            trace_filename: String::from("traceme.ptt"),
             target_pid: unsafe { getpid() },
             tracer_ctx: None,
             data_bufsize: 64,
@@ -104,9 +104,9 @@ impl Tracer {
     /// # Arguments
     ///
     /// * `filename` - The filename in which to store trace packets.
-    pub fn trace_filename(mut self, filename: &str) -> Result<Self, TraceMeError> {
-        self.trace_filename = CString::new(filename)?;
-        Ok(self)
+    pub fn trace_filename(mut self, filename: &str) -> Self {
+        self.trace_filename = String::from(filename);
+        self
     }
 
     /// Select which PID to trace.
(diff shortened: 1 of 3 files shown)


vext01:
Optionally we could flatten `EmptyFileName` into `InvalidFileName`. I was on the fence with this one.

src/lib.rs
```diff
@@ -141,16 +141,18 @@ impl Tracer {
...
-    // Make the map filename by setting/adding a ".map" extension to the trace filename
-    fn make_map_filename(&self, trace_filename: &CString) -> Result<CString, TraceMeError> {
-        let fn_str = trace_filename.clone().into_string()?;
-        let mut pb = PathBuf::from(fn_str);
+    // Make the map filename by setting/adding a ".map" extension to the trace filename.
+    //
+    // # Arguments
+    //
+    //  * `trace_filename` - The trace filename from which to build the map filename.
+    fn make_map_filename(trace_filename: &str) -> Result<String, TraceMeError> {
```
ltratt: Although I know this function isn't public, I tend to think `///` comments make more sense for function definition comments. BTW, I'm not sure the `#Arguments` style is common for Rust (nor am I sure it is very useful for functions with one argument).

src/lib.rs
```diff
@@ -197,3 +203,12 @@ impl Tracer {
         Ok(())
     }
 }
+
+//
```
ltratt: We don't generally bother with comments like this (`#[test]` is an adequate marker IME).

vext01: ok

src/lib.rs
vext01: I just copied the style of Rust by Example: https://rustbyexample.com/meta/doc.html

ltratt: Yes, but the standard docs don't do it, and I can't think of many Crates that do it unthinkingly. Put another way, if you've got lots of complex parameters ("arguments" is more often used to denote the callee site) I think it can make sense to list them out like this; but it's often more pragmatic to detail the 1 or 2 parameters inline in normal prose.

vext01: ok

vext01:
Any thoughts on my comment above about flattening two errors into one?

vext01:
Outstanding comments addressed. One left for you.

vext01:
To expand upon that: `EmptyFileName` could just be a `InvalidFilename(None)`, thus reducing the number of error types we have.

src/lib.rs
```diff
@@ -160,12 +158,16 @@ impl Tracer {
         if self.tracer_ctx.is_some() {
             return Err(TraceMeError::TracerAlreadyStarted);
         }
+        if !self.trace_filename.ends_with(".ptt") {
```
ltratt: I wonder if you really want to use `Path::extension` on this? The nature of filename extensions are OS dependant...

src/lib.rs
```diff
@@ -197,3 +199,8 @@ impl Tracer {
         Ok(())
     }
 }
+
+#[test]
+fn test_make_map_filename_0000() {
```
ltratt: What's the `_0000` for?

ltratt:
Personally, I wouldn't even differentiate between empty and invalid filenames: empty filenames are invalid by definition. So I'd just have `InvalidFilename("")` represent that case.

vext01:
I started making that change, but it's kinda annoying to have to allocate an owned string just for the error. I convinced myself None is the way to go.

src/lib.rs
vext01: It's a unique ID. Imagine we had lots of tests for that function.

ltratt:
Maybe I don't understand how you got to that point without a string...

vext01:
It just complicates the points in the program where you need to make the error, as you will need a `String::From("")`, which is an allocation, whereas `None` is trivially copied. Do you follow?

src/lib.rs
vext01: Like how we did it in Krun:
https://github.com/softdevteam/krun/blob/30d9c6b4d57b497dc3eb13e02ddc72c911223ce2/krun/tests/test_util.py#L139-L230

ltratt:
When you're worried about `"".to_string()` on a branch that will only be called once per execution, I think you've got to the point of worrying too much about performance ;)

vext01:
Meh, I just preferred None, but ok.

vext01:
Ready for me to squash?

ltratt:
The reason that I'm not super keen on None is because then the consumer of the error has to know that None means "empty file name". When you simply sling the string around, the consumer of the error doesn't have to worry about that.

ltratt:
Yep, please squash.

vext01:
splat.

Issue #10, opened 2017-12-08
Decouple from Travis?


ltratt:
At the moment, we have things like `#ifdef TRAVIS` in the code, but I think we can generalise this. It's not exactly Travis that we want to be conditional on, but *any* environment that lacks some of the features we want. As a trivial example, we could change to another CI environment and it would probably have similar restrictions. Or, perhaps, Travis will loosen its restrictions in the future, so what "Travis" means might not even be immutable.

Should we therefore change from "check for platform" to "check for feature"?

vext01:
I think eventually, yes.

I think we should re-visit this later when we are sure that this code will work/be useful in the grand scheme of our tracer.

[Side note: As you pointed out in another PR, we'd need to find out a way to do feature checks with Rust. I'm sure we could figure something out if there is no official way.]

vext01:
I've just found in `/usr/include/linux/perf_event.h` the macros `#define PERF_ATTR_SIZE_VER*` which I think we migh be able to use to decide if the perf implementation is new enough. We might be able to use this to decouple from Travis properly.

ltratt:
An interesting question is whether we should do this at compile-time or run-time. There are obvious advantages and disadvantages to both, but I'm not sure how easy or difficult each might be.

vext01:
The C code won't compile with a too old perf, so compile time it is :)

ltratt:
I guess what I mean is: if we compile with a new perf, can the same binary detect an old perf and disable itself?

vext01:
I'm not sure if that would work, since there are ABI differences. There's a struct for configuring perf which has more members if PT is supported.

ltratt:
If that code isn't called at run-time, it doesn't matter.

Anyway, perhaps this is a red herring. I guess what I *really* care about is "could an OS distribution ship a single binary for a given hardware platform and have it work?" I think the answer is probably yes in either case, but I wouldn't mind being sure!

vext01:
> If that code isn't called at run-time, it doesn't matter.

We indeed use this struct at runtime to configure perf and PT.

> Anyway, perhaps this is a red herring. I guess what I really care about is "could an OS distribution ship a single binary for a given hardware platform and have it work?"

Yes, that's my thinking too. And the answer should be yes.

ltratt:
OK, let's keep an eye on that to make sure that the answer is always "yes" :)

PR #15, opened 2017-12-11
Check we have permission to trace, and fix file permissions.

### LINKED ISSUE #7: Does not need to run as root.
If `/proc/sys/kernel/perf_event_paranoid` contains `-1`, then the tracer can run as an unprivileged user.

vext01:
This change:

 * Checks we have permission to trace.
 * Ensures the `ptt` and `map` files are created with appropriate permissions.

Looks good?

Fixes #7

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,16 +1,20 @@
-use std::ffi;
+use std::{io, ffi, num};
 use std::fmt::{self, Formatter, Display};
+use ::PERF_PERMS_PATH;
 
 #[derive(Debug)]
 pub enum TraceMeError {
     // Wrapped errors from elsewhere.
     FFIIntoString(ffi::IntoStringError),
     FFINul(ffi::NulError),
+    IO(io::Error),
+    NumParseInt(num::ParseIntError),
     // Our own errors.
     CFailure,
     InvalidFileName(String),
     TracerAlreadyStarted,
     TracerNotStarted,
+    TracingNotPermitted,
 }
 
 impl From<ffi::IntoStringError> for TraceMeError {
@@ -25,15 +29,32 @@ impl From<ffi::NulError> for TraceMeError {
     }
 }
 
+impl From<io::Error> for TraceMeError {
+    fn from(err: io::Error) -> Self {
+        TraceMeError::IO(err)
+    }
+}
+
+impl From<num::ParseIntError> for TraceMeError {
+    fn from(err: num::ParseIntError) -> Self {
+        TraceMeError::NumParseInt(err)
+    }
+}
+
 impl Display for TraceMeError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
             &TraceMeError::FFIIntoString(ref e) => write!(f, "{}", e),
             &TraceMeError::FFINul(ref e) => write!(f, "{}", e),
+            &TraceMeError::IO(ref e) => write!(f, "{}", e),
+            &TraceMeError::NumParseInt(ref e) => write!(f, "{}", e),
             &TraceMeError::CFailure => write!(f, "Calling to C failed"),
             &TraceMeError::InvalidFileName(ref n) => write!(f, "Invalid file name: `{}'", n),
             &TraceMeError::TracerAlreadyStarted => write!(f, "Tracer already started"),
             &TraceMeError::TracerNotStarted => write!(f, "Tracer not started"),
+            &TraceMeError::TracingNotPermitted =>
+                write!(f, "Tracing not permitted: you must be root or {} must contain -1",
+                       PERF_PERMS_PATH),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,10 +42,14 @@ extern crate libc;
 
 mod errors;
 
-use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
+use libc::{pid_t, c_char, c_void, getpid, size_t, c_int, geteuid};
 use std::ffi::CString;
 use std::path::PathBuf;
 use errors::TraceMeError;
+use std::fs::File;
+use std::io::Read;
+
+const PERF_PERMS_PATH: &'static str = "/proc/sys/kernel/perf_event_paranoid";
 
 // FFI stubs
 #[link(name = "traceme")]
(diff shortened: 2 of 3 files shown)


vext01:
There's travis sorted.

src/errors.rs
```diff
@@ -25,15 +29,32 @@ impl From<ffi::NulError> for TraceMeError {
...
+    }
+}
+
 impl Display for TraceMeError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
             &TraceMeError::FFIIntoString(ref e) => write!(f, "{}", e),
             &TraceMeError::FFINul(ref e) => write!(f, "{}", e),
+            &TraceMeError::IO(ref e) => write!(f, "{}", e),
+            &TraceMeError::NumParseInt(ref e) => write!(f, "{}", e),
```
ltratt: For all of these cases, you might want to merge them using `|`?

src/lib.rs
```diff
@@ -42,10 +42,14 @@ extern crate libc;
...
 
-use libc::{pid_t, c_char, c_void, getpid, size_t, c_int};
+use libc::{pid_t, c_char, c_void, getpid, size_t, c_int, geteuid};
 use std::ffi::CString;
 use std::path::PathBuf;
 use errors::TraceMeError;
+use std::fs::File;
+use std::io::Read;
+
+const PERF_PERMS_PATH: &'static str = "/proc/sys/kernel/perf_event_paranoid";
```
ltratt: I think this should be guarded with a `#[cfg(Linux)]` or whatever it is. At some point we'll want this run on non-Linux, so having a compile-time error will make it easier to port.

src/errors.rs
vext01: Will it work? The `e`s have different types.

src/lib.rs
```diff
@@ -151,10 +155,31 @@ impl Tracer {
...
+        if unsafe { geteuid() } == 0 {
+            // Root can always trace.
+            return Ok(());
+        }
+
+        let mut f = File::open(&PERF_PERMS_PATH)?;
+        let mut buf = String::new();
+        f.read_to_string(&mut buf)?;
+        let buf = buf.trim();
+        let perm = buf.parse::<i8>()?;
```
ltratt: What's the use of having `perm` is an intrmediate variable? Wouldn't `if buf.trim().parse::<i8>()? != -1 { return Err(...) }` be simpler?

src/lib.rs
vext01: Shall we just say this doesn't work on !Linux in the README? The whole thing relies on perf. Or put on test at an entry point or something?

src/lib.rs
vext01: Can do, yeah.

src/errors.rs
ltratt: Ah, I hadn't though of that. They probably won't work... but would be interesting to try!

src/lib.rs
vext01: Let's make build.rs crash on !Linux.

ltratt: From experience of porting, it's *much* better if you make each individual point of likely divergence fail. Just saying "it doesn't work on !Linux" globally doesn't help a porter much.

src/errors.rs
vext01: They all implement `Error`. So maybe...

ltratt: If `Error` implements the appropriate formatter, it will probably work I think.

src/lib.rs
vext01: OK, but they will be everywhere.

src/errors.rs
vext01: No cigar.

```rust
        match self {    
            &TraceMeError::FFIIntoString(ref e)                           
            | &TraceMeError::FFINul(ref e)                           
            | &TraceMeError::IO(ref e)                                          
            | &TraceMeError::NumParseInt(ref e) => write!(f, "{}", e),    
```

```
error[E0308]: mismatched types
  --> src/errors.rs:48:37
   |
48 |             | &TraceMeError::FFINul(ref e)
   |                                     ^^^^^ expected struct `std::ffi::IntoStringError`, found struct `std::ffi::NulError`
   |
   = note: expected type `&std::ffi::IntoStringError`
              found type `&std::ffi::NulError`

error[E0308]: mismatched types
  --> src/errors.rs:49:33
   |
49 |             | &TraceMeError::IO(ref e)
   |                                 ^^^^^ expected struct `std::ffi::IntoStringError`, found struct `std::io::Error`
   |
   = note: expected type `&std::ffi::IntoStringError`
              found type `&std::io::Error`

error[E0308]: mismatched types
  --> src/errors.rs:50:42
   |
50 |             | &TraceMeError::NumParseInt(ref e) => write!(f, "{}", e),
   |                                          ^^^^^ expected struct `std::ffi::IntoStringError`, found struct `std::num::ParseIntError`
   |
   = note: expected type `&std::ffi::IntoStringError`
              found type `&std::num::ParseIntError`

error: aborting due to 3 previous errors
```

Also tried annotating `e`. You can't do that it seems.

src/lib.rs
vext01: In fact, I'd like to do that as a separate PR, if that's OK?

ltratt: Well if they end up in too many places, maybe we need to rethink the code layout a bit?

ltratt: I've no problem with that being a separate PR.

src/errors.rs
ltratt: OK, sorry for the noise!

src/lib.rs
vext01: I think we should ease off the software engineer a little for now, since we have no idea if our ideas will work yet. I'm very keen to move on and it doesn't need to be rock solid just yet. We can introduce platform abstractions later.

ltratt: Point taken. But keep it in mind: it's really painful to introduce these abstractions to even a moderately sized project. So it might be better to bite the bullet at some point not ridiculously far into the future.

vext01: #17.

You might be right actually. So I think we would want switchable backends, maybe a Travis backend too, which would somewhat address somewhat #10. Hrm. The perf backend basically consists of all of the current C code, so I'd even be tempted to rename traceme.c to `backend_perf.c` or something.

Let me integrate this into Rustc, then we can come back to this.

vext01:
So I think that's all the comments addressed.

ltratt:
OK, if you squash, I can merge.

vext01:
Squashed.

Issue #17, opened 2017-12-11
Platform abstractions.


vext01:
In the long term, we'd like this to work on other systems than linux/perf. Maybe even linux/notperf.

In light of that, we should add platform abstraction before we get too deeply involved.

vext01:
Ideas:

 * Make different backends using Traits.
 * Consider Travis a backend of it's own.
 * The backend we are using now can become the `perf backend`

ltratt:
I suspect conditional compilation, including conditional compilation of modules, will also prove a very effective tactic. In a sense, best practise in making portable C code probably isn't *too* far off the mark here either.

vext01:
Also fixed. See #19

Issue #16, opened 2017-12-11
Make linux specific code crash on other platforms.


vext01:

ltratt:
By "crash", we ideally mean "refuses to compile, stating clearly which bit won't work on another platform".

vext01:
yes, crash at compile time.

vext01:
#19

Issue #8, opened 2017-12-06
Trace to memory buffer.


vext01:
Currently the trace goes to a file. Eventually, once we understand more, we will need to trace to a buffer and decode it in parallel with the interpreter.

vext01:
This is one of the next tasks on my list.

I think the tracer backends should include a way to recieve "chunks" of the trace incrementally, probably via a channel. This should be built into the `Tracer` trait so that it can be done in a generic fashion for all backends.

vext01:
This is next on the list.

The code that reads trace data out of the `mmap`ed buffer is currently written in C, and runs in a thread:
https://github.com/softdevteam/traceme/blob/a932bec22f7e59495d770f59b409b9bdcc7cf799/src/backends/perf_pt/perf_pt.c#L195

We need to think about how to get this out of C, and into rust.

Goodish Ideas:
 1. Pipe 
 2. Have the C code heap allocate and have an accessor method to get at the raw pointer once we are done tracing.
 3. Try to translate the poll loop into Rust code and use a channel.

`1` would work, but might introduce delays.

`2` is good, but would disallow "streaming" the trace. We might like to process the trace in parallel with it being collected?

`3` is good, but would require some (hopefully fairly light) architectural changes (spawn the thread in Rust, not C).

I'm tempted to try 3.

Do you have any thoughts @ltratt 

[Bad ideas i'm not conisidering are: named pipes, files, shared memory, mongodb^W :smile: )

ltratt:
Quick thoughts: no way to pipes (*massive* overheads, relatively speaking); I don't mind 2 or 3 too much. I'm not sure we're going to be able to stream the data in practise, and certainly not at first. So if 3 is very hard, I'm OK with 2.

vext01:
OK, let me expand on `2` a moment then.

What I'd do is start by allocating a buffer as big as the trace buffer itself, and empty the contents of the PT buffer as the poll loop is awoken. If the buffer becomes full, I'd probably reallocate it at twice the capacity and continue.

Once the trace is finished, we'd need a `tracer.trace_buf() -> *const c_void` or similar method which returns the address of the buffer. This will eventually be sent off to a packet decoder (probably a binding to libipt (https://github.com/01org/processor-trace) so we can reconstruct the instruction stream.

`2` is easier than `3`. The only nit is that we will have to provide a method to free the trace buffer once we are done.

Still OK for me to go ahead with `2`?

ltratt:
[...]

PR #19, opened 2018-01-02
Introduce "tracing backends" into traceme.

### LINKED ISSUE #10: Decouple from Travis?

vext01:
To begin with we have "pt_perf" and "dummy" backends.

 * pt_perf uses Linux Perf to collect Intel Processor Trace packets.
 * dummy is a fall-back backend which does nothing.

Conditional compilation is used to include/exclude the pt_perf backend
depending upon the the outcome of a (crude) feature check in build.rs.

It's up to the consumer if traceme to choose an appropriate tracing backend and configure it (instantiation and configuration is specific to each tracer, and not part of the `Tracer` trait).

Note that the compiler output appears for failing feature checks appear in the `cargo build` output (and so in the travis log too). A bit annoying, but not a show stopper. I'm hoping the Rust community will devise a better feature check mechanism in the future.

Looks good?

Fixes #10, #16, #17.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -36,41 +36,29 @@
 // SOFTWARE.
 
 extern crate gcc;
-use std::env;
-use std::process::Command;
-use std::collections::HashMap;
+use std::path::PathBuf;
 
-fn run_lsb_release(args: Vec<&str>) -> HashMap<String, String> {
-    let output = Command::new("lsb_release")
-                         .args(args)
-                         .output()
-                         .unwrap();
-    assert!(output.status.success());
-    let output_s = String::from_utf8(output.stdout).unwrap();
+const FEATURE_CHECKS_PATH: &str = "feature_checks";
 
-    let mut res = HashMap::new();
-    for line in output_s.lines() {
-        let parts: Vec<&str> = line.split(":").collect();
-        assert!(parts.len() == 2);
-        res.insert(String::from(parts[0].trim()), String::from(parts[1].trim()));
-    }
-    res
+/// Simple feature check, returning `true` if we have the feature.
+///
+/// The checks themselves are in files under `FEATURE_CHECKS_PATH`.
+fn feature_check(filename: &str) -> bool {
+    let mut path = PathBuf::new();
+    path.push(FEATURE_CHECKS_PATH);
+    path.push(filename);
+
+    let mut check_build = gcc::Build::new();
+    check_build.file(path).try_compile("check_perf_pt").is_ok()
 }
 
 fn main() {
     let mut c_build = gcc::Build::new();
-    c_build.file("src/traceme.c").include("src");
 
-    // Travis' Linux setup is too old for what we do with linux-perf.
-    if env::var("TRAVIS").is_ok() {
-        // Crash out when Travis updates to a newer Ubuntu or a different OS.
-        // When this fires, travis may have a new enough linux-perf to test properly.
-        let lsb_rel = run_lsb_release(vec!["-ir"]);
-        assert!(lsb_rel["Distributor ID"] == "Ubuntu");
-        assert!(lsb_rel["Release"] == "14.04");
-        // Setting -DTRAVIS causes the C API to be stubbed.
-        c_build.define("TRAVIS", None);
-        println!("cargo:rustc-cfg=travis");
+    // Check for a new enough Perf to support Intel PT.
+    if feature_check("check_perf_pt.c") {
+        c_build.file("src/backends/perf_pt/perf_pt.c");
+        println!("cargo:rustc-cfg=perf_pt");
     }
 
     c_build.compile("traceme");
--- a/src/backends/dummy.rs
+++ b/src/backends/dummy.rs
@@ -35,52 +35,56 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-extern crate traceme;
-extern crate libc;
+use Tracer;
+use errors::TraceMeError;
 
-use traceme::Tracer;
-use libc::getpid;
+/// A tracer which doesn't really do anything.
+pub struct DummyTracer {
+    started: bool,
+}
 
-use traceme::backends::DummyTracer;
-#[cfg(all(perf_pt, target_arch = "x86_64"))]
-use traceme::backends::PerfPTTracer;
-
-/// Instantiate a tracer suitable for the current platform.
-fn tracer() -> Box<Tracer> {
-    if cfg!(target_os = "linux") {
-        #[cfg(all(perf_pt, target_arch = "x86_64"))] {
-            let res = PerfPTTracer::new();
-            if res.is_ok() {
-                return Box::new(res.unwrap().trace_filename("simple_example.ptt"));
-            } else {
-                // CPU doesn't have Intel PT support.
-                return Box::new(DummyTracer::new());
-            }
-        }
-        #[cfg(not(all(perf_pt, target_arch = "x86_64")))] {
-            return Box::new(DummyTracer::new());
-        }
-    } else {
-        return Box::new(DummyTracer::new());
+impl DummyTracer {
+    /// Create a dummy tracer.
+    pub fn new() -> Self {
+        Self {started: false}
     }
 }
 
-/// Trace a simple computation loop.
-///
-/// The result is printed to discourage the compiler from optimising the computation out.
-fn main() {
-    let mut res = 0;
-    let pid = unsafe { getpid() };
-
-    let mut tracer = tracer();
-    tracer.start_tracing().unwrap_or_else(|e| {
-        panic!(format!("Failed to start tracer: {}", e));
-    });
-
-    for i in 1..10 {
-        res += i + pid;
+impl Tracer for DummyTracer {
+    fn start_tracing(&mut self) -> Result<(), TraceMeError> {
+        if self.started {
+            return Err(TraceMeError::TracerAlreadyStarted);
+        }
+        self.started = true;
+        Ok(())
     }
 
-    tracer.stop_tracing().unwrap();
-    println!("result: {}", res);
+    fn stop_tracing(&mut self) -> Result<(), TraceMeError> {
+        if !self.started {
+            return Err(TraceMeError::TracerNotStarted);
+        }
+        self.started = false;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::DummyTracer;
+    use ::test_helpers;
+
+    #[test]
+    fn test_basic_usage() {
+        test_helpers::test_basic_usage(DummyTracer::new());
+    }
+
+    #[test]
+    fn test_already_started() {
+        test_helpers::test_already_started(DummyTracer::new());
+    }
+
+    #[test]
+    fn test_not_started() {
+        test_helpers::test_not_started(DummyTracer::new());
+    }
 }
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -1,0 +1,284 @@
+// Copyright (c) 2017 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use libc::{pid_t, c_char, c_void, getpid, size_t, c_int, geteuid};
+use std::ffi::CString;
+use std::path::PathBuf;
+use errors::TraceMeError;
+use std::fs::File;
+use std::io::Read;
+use Tracer;
+
+// The sysfs path used to set perf permissions.
+const PERF_PERMS_PATH: &'static str = "/proc/sys/kernel/perf_event_paranoid";
+
+// FFI prototypes.
+extern "C" {
+    fn perf_pt_start_tracer(conf: *const PerfPTConf) -> *const c_void;
+    fn perf_pt_stop_tracer(tr_ctx: *const c_void) -> c_int;
+}
+
+// Struct used to communicate a tracing configuration to the C code. Must
+// stay in sync with the C code.
+#[repr(C)]
+struct PerfPTConf {
+    target_pid: pid_t,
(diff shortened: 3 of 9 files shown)


vext01:
Any comments for this one?

ltratt:
Oops! This one passed me by -- sorry!

build.rs
```diff
@@ -36,41 +36,29 @@
...
-        res.insert(String::from(parts[0].trim()), String::from(parts[1].trim()));
-    }
-    res
+/// Simple feature check, returning `true` if we have the feature.
+///
+/// The checks themselves are in files under `FEATURE_CHECKS_PATH`.
+fn feature_check(filename: &str) -> bool {
+    let mut path = PathBuf::new();
+    path.push(FEATURE_CHECKS_PATH);
+    path.push(filename.clone());
```
ltratt: I don't think you need the `clone` here?

examples/simple_example.rs
```diff
@@ -41,14 +41,35 @@ extern crate libc;
 use traceme::Tracer;
 use libc::getpid;
 
+use traceme::backends::DummyTracer;
+#[cfg(perf_pt)]
+use traceme::backends::PerfPTTracer;
+
```
ltratt: Double blank line.

examples/simple_example.rs
```diff
@@ -41,14 +41,35 @@ extern crate libc;
...
+
+/// Instantiate a tracer suitable for the current platform.
+fn init_tracer() -> Box<Tracer> {
+    if cfg!(target_os = "linux") {
+        #[cfg(perf_pt)] {
+            return Box::new(PerfPTTracer::new().trace_filename("simple_example.ptt"));
+        }
+        #[cfg(not(perf_pt))] {
+            println!("Warning: perf_pt backend not compiled in. Using dummy tracer backend.");
```
ltratt: I don't see the point in the warning: printing this out at runtime isn't really going to help users.

examples/simple_example.rs
```diff
@@ -41,14 +41,35 @@ extern crate libc;
...
+            return Box::new(DummyTracer::new());
+        }
+    } else {
+        println!("Warning: unsupported platform. Using dummy tracer backend.");
```
ltratt: Same comment as above.

src/backends/dummy.rs
```diff
@@ -0,0 +1,61 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use Tracer;
+use errors::TraceMeError;
+
+/// A tracer which doesn't really do anything.
+///
+/// This exists for (limited) testing of Travis CI.
```
ltratt: I don't think we should say this is *just* for Travis. It could be useful elsewhere right?

src/backends/perf_pt/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// # Example
+    ///
+    /// ```
+    /// use traceme::backends::PerfPTTracer;
+    ///
+    /// PerfPTTracer::new().trace_filename("mytrace.ptt").data_bufsize(1024).target_pid(666);
+    /// ```
+    pub fn new() -> Self {
+        Self {
+            trace_filename: String::from("traceme.ptt"),
```
ltratt: Having a fixed filename feels weird to me. Should it be configurable?

src/backends/perf_pt/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+        self
+    }
+
+    /// Make the map filename by setting/adding a ".map" extension to `trace_filename`.
+    fn make_map_filename(trace_filename: &str) -> Result<String, TraceMeError> {
+        let mut pb = PathBuf::from(trace_filename);
+        if !pb.set_extension("map") {
+            return Err(TraceMeError::InvalidFileName("".to_string()));
+        }
+        let map_filename = pb.to_str().ok_or(TraceMeError::InvalidFileName("".to_string()))?;
```
ltratt: Why do we say there's an InvalidFileName of an empty string?

src/backends/perf_pt/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl Tracer for PerfPTTracer {
+    fn start_tracing(&mut self) -> Result<(), TraceMeError> {
+        PerfPTTracer::check_perf_perms()?;
+        if self.tracer_ctx.is_some() {
```
ltratt: Shouldn't this `if` be a statement ending in `?`?

src/lib.rs
```diff
@@ -40,192 +40,23 @@
...
-    /// Create a new tracer.
-    ///
-    /// # Example
-    ///
-    /// ```
-    /// use traceme::Tracer;
+/// The interface offered by all tracer types.
+///
+/// It is the job of the consumer of this library to decide which specific tracer type to
+/// instantiate, and to set it up using its type-specific configuration methods. Once configured a
```
ltratt: Missing comma after "Once configured".

src/lib.rs
```diff
@@ -40,192 +40,23 @@
...
+/// tracer can be treated as a generic data structure with Box<Tracer>.
```
ltratt: I don't think we need to mention the `Box` thing here (plus the phrasing doesn't currently make sense). It's up to users whether they want dynamic dispatch on a `Tracer` or not.

ltratt:
I need to take another look at this to think about the architecture (will do tomorrow), but most of the low-level details look reasonable.

Two quick comments. 1) do things like a default filename make sense? I worry that API users might think such a default is OK when I think the intention is that users really must provide their own filename (in which case it should go in the constructuor I guess?). 2) I think that having markdown parameters spread over multiple lines in doc really takes up vast acres of (precious) vertical space and should be removed (the standard docs don't use this idiom). I had nightmares of javadoc ;)

examples/simple_example.rs
```diff
@@ -41,14 +41,35 @@ extern crate libc;
 use traceme::Tracer;
 use libc::getpid;
 
+use traceme::backends::DummyTracer;
+#[cfg(perf_pt)]
+use traceme::backends::PerfPTTracer;
+
+
+/// Instantiate a tracer suitable for the current platform.
+fn init_tracer() -> Box<Tracer> {
```
ltratt: I think this function name is misleading. It doesn't initialise a tracer: it makes a new one. So maybe just call it `tracer`?

ltratt:
I think my major architectural question is whether dynamic dispatch is necessary. If you think about `init_tracer` it currently returns a single type determined at compile-time. Put another way, any given binary only has 1 tracer type in it, so can't really benefit from dynamic dispatch.

Now, if we only call into the `Box<Tracer>` infrequently, this doesn't really matter either way. If we call into it regularly, the dynamic dispatch will start to hurt. I'm not sure if we will be calling into it regularly or not though. Any thoughts?

build.rs
vext01: correct.

examples/simple_example.rs
vext01: This was intentional to separate imports from functions. Nevertheless, if thats the style we prefer, ok.

examples/simple_example.rs
vext01: I printed it because, well you don't really test anything in the case that you use the dummy backend.

How about we instead do the Rust equivalent of:
```python
def test_dummy():
    # test always run
    ...

@pytest.skipif(not supports_perf_pt())
def test_pt_perf():
    # only run if we suport pt_perf
    ...
```

i.e. test each backend separately? Then I'm happy to kill the warning (and the `init` function for that matter too).

src/backends/dummy.rs
vext01: Oops, there's supposed to be an 'e.g.' in there somewhere.

vext01: Yes, it would be useful to allow the interpreter to at least run on platforms where there is no JIT support.

src/backends/perf_pt/mod.rs
vext01: It is configurable, yes:
https://github.com/softdevteam/traceme/blob/678ff424831a866e760fd4b242278e24e8318222/src/backends/perf_pt/mod.rs#L101-L109

However, don't worry too much about the trace filenames. I'm going to tear this out when I work on #8. We don't want to trace to disk, as it will be too slow.

[I might add the ability to dump traces to disk as a debug facility only (and the i'd derive the trace filename from the pid and the address of the `Location`)]

src/backends/perf_pt/mod.rs
vext01: PathBuf::set_extension:
> Returns false and does nothing if self.file_name is None, returns true and updates the extension otherwise.

So if we get `false` there's an empty filename.

We discussed in #13 whether an empty filename should be a distinct error, or an `InvalidFileName` with an empty string, and the latter came up on top:

> ltratt: Personally, I wouldn't even differentiate between empty and invalid filenames: empty filenames are invalid by definition.

https://github.com/softdevteam/traceme/pull/13#issuecomment-350702887

This code will disappear when we use the linker API to get the maps, see #20.

src/backends/perf_pt/mod.rs
vext01: You mean using an `and*` combinator or something?

src/lib.rs
vext01: OK, I guess you are right. Let's assume the user knows that they can't store a trait object without an additional wrapper like a `Box`.

examples/simple_example.rs
vext01: Sure, but it might disappear, see the comment above about tests:
https://github.com/softdevteam/traceme/pull/19#discussion_r160356429

vext01:
> I think my major architectural question is whether dynamic dispatch is necessary.

> If you think about init_tracer it currently returns a single type determined at compile-time. Put another way, any given binary only has 1 tracer type in it, so can't really benefit from dynamic dispatch.

OK, so I have thought long and hard about this. Here's my thoughts.

Let's first roll with your assumption that each platform has only one backend. To avoid dynamic dispatch in the user's interpreter, they would have something like:

```rust
struct MyVM {
    #[cfg(target_os = "linux")]
    tracer: PerfPTTracer,
    #[cfg(target_os = "openbsd")]
    tracer: OpenBSDTracer,
    ...
}

impl MyVM {
    pub fn init() {
        #[cfg(target_os = "linux")] {
            // call PerfPTTracer specific config methods
            ...
        }

        #[cfg(target_os = "openbsd")] {
             // call OpenBSDTracer specific config methods
            ...
        }
    ...
}
```

So you would use conditional compilation to store the right tracer type and make sure you use the right configuration methods.

[I'm 100% convinced you can't get away from the platform specific configuration in `init()`. Even using a "property-based" configuration API where all tracers have the same config functions, you are still going to want to call configuration functions with different arguments depending on the platform.]

But wait, we already have a platform where we have more than one backend, so either:
 1. The user must bake one backend in at compile time.
 2.  The user can choose at runtime which to use.

Which would we want? The latter (the model I've been assuming) lends itself to dynamic dispatch, although it would be possible to avoid it by having multiple struct members in `MyVM`, one for each possible backends on that platform (at the cost of a little more memory consumption).

TBC...

vext01:
> Now, if we only call into the Box<Tracer> infrequently, this doesn't really matter either way.

This is an interesting question.

The perf_event api currently requires one call to `perf_event_open()` for each pid. Under linux, each thread is a distinct PID. So in other words, we have to open one fd for each new thread we want to trace. Therefore it depends on the workload of the user program as to how frequently we make a new tracer.

```
int perf_event_open(struct perf_event_attr *attr,
                           pid_t pid, int cpu, int group_fd,
                           unsigned long flags);
```

But now I wonder if we want to have *one* tracer instance for all threads? i.e. have a collection of perf fds inside a `PerfPTTracer` instance. Then you don't need to make a new instance for each new thread, thus `Box` becomes less of a problem.

More thought needed! Let's talk in person today.

examples/simple_example.rs
ltratt: Yep, I understand that we can't test things in such cases. I tend to agree that some sort of "don't test things that can't be tested" dummy is a sensible sounding idea.

src/backends/perf_pt/mod.rs

src/backends/perf_pt/mod.rs

src/backends/perf_pt/mod.rs
ltratt: Yes, something like that.

ltratt:
If we can only have one valid tracer, you could have:

```rust
struct MyVM {
    tracer: Tracer,
    ...
}
```

and then:

```rust
#[cfg(target_os = "linux")]
fn tracer() -> Tracer {
   PerfTracer::new()
}

#[cfg(target_os = "openbsd")]
fn tracer() -> Tracer {
   DummyTracer::new()
}
```

You'd only need dynamic dispatch, as far as I can see, if you could have more than 1 tracer.

vext01:
Linux needs both `DummyTracer` (for travis for now) and `PerfPTTracer`. It's also not inconceivable that we might write our own PT backend for Linux, whilst keeping the PT backend in-tree. So I think we should allow each platform to have >1 backend.

[maybe related: I didn't know you could store a trait object in a struct like that.]

vext01:
> [maybe related: I didn't know you could store a trait object in a struct like that.]

https://play.rust-lang.org/?gist=83cdeec280b72040674344a0dcc6f79b&version=stable

The problem is, you can't store trait objects on the stack since the concrete size is unknown. You'd need a `Box` (or similar?) around it.

Is there a way around that?

vext01:
For the PR history, we've agreed that it's OK to use a trait with the box.

examples/simple_example.rs
vext01: So, does that mean, "yes, let's test the two backends separately"?

ltratt: We could test the dummy backend *always*, and only test the PT backend if it's enabled?

vext01: Yep, that's exactly what I had in mind. See the example listing above.

src/backends/perf_pt/mod.rs
vext01: I'm not sure these help actually. We have:

```
fn and<U>(self, optb: Option<U>) -> Option<U>[src]

Returns None if the option is None, otherwise returns optb.
```

 and

```
fn and_then<U, F>(self, f: F) -> Option<U> where

Returns None if the option is None, otherwise calls f with the wrapped value and returns the result.
```

We would need something which returns a `Result`.

We do have:
```
fn ok_or<E>(self, err: E) -> Result<T, E>[src]

Transforms the Option<T> into a Result<T, E>, mapping Some(v) to Ok(v) and None to Err(err).
```

But we would want the opposite mapping.

Am I missing something?

examples/simple_example.rs
ltratt: Then we're in agreement :)

src/backends/perf_pt/mod.rs
ltratt: Ignore this comment -- I was being an idiot. Mea culpa!

vext01:
Alright, I think this is ready for re-review now.

[Sorry it took so long. LLVM inline asm is a slightly different dialect to GNU, which led to some subtle bugs]

There are a couple of nits:
 * The map file permissions code was wrong. I've kludged it for now. This code will be killed entirely shortly (#20).
 * We have to run tests sequentially until we implement #20, otherwise the tests race the map file and hilarity ensues.

Other than that, I'm pretty happy with this. Are you?

src/lib.rs
```diff
@@ -59,3 +59,35 @@ pub trait Tracer {
...
+    pub fn test_basic_usage<T>(mut tracer: T) where T: Tracer {
+        tracer.start_tracing().unwrap();
+        tracer.stop_tracing().unwrap();
+    }
+
+    /// Check that starting a tracer twice makes an appropriate error.
+    pub fn test_already_started<T>(mut tracer: T) where T: Tracer {
+        tracer.start_tracing().unwrap();
+        match tracer.start_tracing() {
+            Err(TraceMeError::TracerAlreadyStarted) => { },
```
ltratt: It's more common to say `=> ()` I think.

src/backends/perf_pt/perf_pt.c
```diff
@@ -137,10 +137,9 @@ stash_maps(pid_t pid, const char *map_filename)
     DEBUG("saving map to %s", map_filename);
     bool ret = true;
 
-    mode_t old_mode = umask(MAPS_MODE);
-
     char *cmd = NULL;
-    int res = asprintf(&cmd, "cp /proc/%d/maps %s", pid, map_filename);
+    int res = asprintf(&cmd, "cp /proc/%d/maps %s && chmod 600 %s",
```
ltratt: Ouch, getting the shell to deal with filenames is a security disaster waiting to happen (e.g. filenames with spaces). Unless we're 100% sure this code will be entirely killed in the next PR, I think we have to escape things properly.

ltratt:
OK, I have one mnior and one minor/major (who knows) comment.

vext01:
Yep. This code is garbage. I kludged it just to get tests working.

I'm very sure I will delete this whole function soon. There's an API to get
the maps, which avoids faffing with files entirely.

ltratt:
OK, please make that the next PR. That code is so dangerous, it scares me!

vext01:
You have my word on that!

ltratt:
Glad to hear :) OK, then we have that one style comment... then I think we'll be ready to merge.

vext01:
OK to do some squashing?

ltratt:
Yep, please squash!

vext01:
splat.

(Hacks in separate commits)

PR #21, opened 2018-01-12
Kill map file

vext01:
This kills the "stashing of the map file" and the crufty code around it.

Instead of copying a file out of procfs (which relies on the OS having proc, and would have to be sucked in and parsed later), we use the semi-portable `dl_iterate_phdr()` API to work with in-memory data structures. This also helps with tests, since the map flie is one less file to race in parallel testing (the trace file still races -- fixing in #8).

The new code isn't used (apart from in tests), but will be needed later to relate PT packets to the binary code.

Looks good?

### DIFF
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -1,0 +1,108 @@
+// Copyright (c) 2018 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use libc::{c_int, uintptr_t};
+use ::errors::TraceMeError;
+
+// FFI prototypes.
+extern "C" {
+    fn traceme_exec_base(addr: *const uintptr_t) -> c_int;
+}
+
+/// Get the relocated virtual start address of the executable code for the current program.
+pub fn exec_base() -> Result<usize, TraceMeError> {
+    let mut addr: uintptr_t = 0;
+    match unsafe {traceme_exec_base(&mut addr as *const uintptr_t)} {
+        0 => Err(TraceMeError::ElfError(String::from("Failed to get executable base address"))),
+        1 => Ok(addr as usize),
+        _ => unreachable!(),
+    }
+}
+
+#[cfg(all(test, target_os = "linux"))]
+mod test_linux {
+    use super::exec_base;
+
+    use libc::getpid;
+    use std::path::PathBuf;
+    use std::fs::File;
+    use std::io::BufReader;
+    use std::io::BufRead;
+    use std::env::current_exe;
+
+    /// Check we can get the exec_base() address and that the procfs map file agrees.
+    #[test]
+    fn test_linux_exec_base() {
+        let got_base = exec_base().unwrap();
+
+        let mut path = PathBuf::from("/proc");
+        path.push(unsafe {getpid()}.to_string());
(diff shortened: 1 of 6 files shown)


vext01:
[Happy to pre-squash if you'd prefer]

src/util/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+        1 => Ok(addr as usize),
+        _ => unreachable!(),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::exec_base;
+
+    #[cfg(target_os = "linux")]
```
vext01: The repeated platform checks here annoy me.

I was hoping to use:
```
#[cfg(target_os = "linux")] {
    ...
}
```

But you can't :(

ltratt:
Yes, please do.

vext01:
Try this.

src/util/util.c
```diff
@@ -0,0 +1,108 @@
...
+
+    // Check if this is the entry for the binary itself (not a shared object).
+#if defined(__linux__)
+    // On Linux the main binary has no name.
+    if (info->dlpi_name[0] != '\0') {
+        return 0;
+    }
+#elif defined(__OpenBSD__)
+    // On OpenBSD the name is that of the binary.
+    if (strcmp(basename(info->dlpi_name), basename(getprogname())) != 0) {
```
vext01: Oops, too many basenames here. Didn't notice before. Fix coming.

ltratt:
Ready for rereview, or is that new commit going to be squashed first or ...?

vext01:
Here we go.

src/util/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+use ::errors::TraceMeError;
+
+// FFI prototypes.
+extern "C" {
+    fn traceme_exec_base(addr: *const uintptr_t) -> c_int;
+}
+
+/// Get the relocated virtual start address of the executable code for the current program.
+pub fn exec_base() -> Result<usize, TraceMeError> {
+    let mut addr: uintptr_t = 666;
```
ltratt: Do we need to give it a default value? If we do, let's choose something that won't offend any people (like 0...).

src/util/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    match unsafe {
```
ltratt: I would probably put this `unsafe` block on one line myself.

src/util/mod.rs
ltratt: Indeed, that would scope them.

ltratt: Though you could, I think, do:

```rust
#[cfg(target_os = "linux")] {
  use <linux stuff>;

  #[test]
  fn test_linux1 {...}
  fn test_linux2 {...}
}
```

src/util/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[cfg(target_os = "linux")]
+    use std::io::BufRead;
+    #[cfg(target_os = "linux")]
+    use std::env::current_exe;
+
+    /// Check we can get the exec_base() address.
+    ///
+    /// On Linux, also check this agrees with the procfs map file.
+    #[test]
+    fn test_exec_base() {
```
ltratt: This feels like it should be one test for Linux and another for not-Linux.

src/util/mod.rs
vext01: Rust insists, but 0 is ok.

src/util/mod.rs
vext01: Can do.

src/util/util.c
```diff
@@ -0,0 +1,108 @@
...
+// Exposed prototypes.
+int traceme_exec_base(uintptr_t *);
+
+/*
+ * Called by `dl_iterate_phdr` once for each program header.
+ *
+ * We are looking for the start address of the program's executable
+ * instructions.
+ */
+static int
```
ltratt: Is the return type an `int` or a `bool`?

src/util/util.c
```diff
@@ -0,0 +1,108 @@
...
+    ElfW(Phdr) phdr;
+    ElfW(Half) i;
+#endif
+    (void) size;
+
+    // Check if this is the entry for the binary itself (not a shared object).
+#if defined(__linux__)
+    // On Linux the main binary has no name.
+    if (info->dlpi_name[0] != '\0') {
+        return 0;
```
ltratt: Hang on, `0` is the success return code I assume, but it seems to being returned here on failure?

src/util/util.c
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+    return 0;
+}
+
+/*
+ * Search program headers for the relocated start address of the current
+ * program's executable code.
+ *
+ * Returns 1 if the address is found, or 0 otherwise. If found, the address is
```
ltratt: Let's just make it return a bool and be done with it.

src/util/mod.rs
vext01: Could do, but one would be a subset of the other.

ltratt: That's fine. I think it would be better than having conditional compilation inside the function (which is easily misread IMHO).

src/util/util.c
vext01: Nice try, but the callback has to return an int I'm afraid:
```
     int
     dl_iterate_phdr(int (*callback)(struct dl_phdr_info *, size_t, void*),
         void *data);
```

ltratt: Ah, can we make it clear (even very briefly) that the API is forced on us by `dl_iterate_phdr()` (I missed that).

src/util/util.c
vext01: This is the interface required by the `dl_iterate_phdr` interface.

```
      The dl_iterate_phdr() function walks through the list  of  an  applica‐
       tion's  shared  objects  and  calls the function callback once for each
       object, until either all shared objects have been processed or callback
       returns a nonzero value.
```

We can invert the logic in `traceme_exec_base` if you like though.

Shall I move the explanation?

src/util/mod.rs
vext01: ok

src/util/util.c
ltratt: Maybe move `traceme_exec_base` above this function (assuming that most people read from top to bottom!)?

src/util/util.c
vext01: No, we still can't pass bools over the rust<->c boundary. Sorry :(

src/util/util.c
vext01: Sure thing!

src/util/mod.rs
vext01:
```
    #[cfg(target_os = "linux")] {    
        use libc::getpid;                                                          
        use std::path::PathBuf;                         
        use std::fs::File;         
        use std::io::BufReader;    
        use std::io::BufRead;                        
        use std::env::current_exe;    
    }
```

Will give:

```
error: expected item after attributes
  --> src/util/mod.rs:62:31
   |
62 |     #[cfg(target_os = "linux")] {
```

Because `use` is not an `item`.

The solution I came up with uses a `mod` for linux specific tests, and another for generic tests.

```
#[cfg(all(test, target_os = "linux"))]
mod test_linux { 
    ...
}

#[cfg(test)]
mod test { 
    ...
}
```

How do you find that?

vext01:
Ready for re-review.

ltratt:
OK, I think we're ready to squash.

vext01:
splat.

PR #24, opened 2018-01-18
Fix a bug where perf_pt tests don't receive any trace data.

vext01:
For the perf_pt backend, we want to trace a thread ID (TID), not a
process (PID). A TID is a Linux specific notion, and unfortunately pid_t
is re-used for TIDs, which can lead to some confusion. I've updated doc
strings to try and make it clear to users.

The example code worked because the first thread of a program has the same TID as the PID. `cargo test` however, spawns a thread (with a new TID). We were tracing the main thread that's sat there twiddling it's thumbs waiting for the test thread to exit.

We can see the issue by looking at the debug output before and after this fix.

Before:
```
test backends::perf_pt::tests::test_basic_usage ...
...
rc/backends/perf_pt/perf_pt.c:220 [poll_loop]: wake
src/backends/perf_pt/perf_pt.c:221 [poll_loop]: aux_head=  0x0000000000
src/backends/perf_pt/perf_pt.c:222 [poll_loop]: aux_tail=  0x0000000000
src/backends/perf_pt/perf_pt.c:223 [poll_loop]: aux_offset=0x0000266240
src/backends/perf_pt/perf_pt.c:224 [poll_loop]: aux_size=  0x0004194304
```

Notice the head and tail are the same, so we read no data.

After:
```
test backends::perf_pt::tests::test_basic_usage ...
...
src/backends/perf_pt/perf_pt.c:220 [poll_loop]: wake
src/backends/perf_pt/perf_pt.c:221 [poll_loop]: aux_head=  0x0000000944
src/backends/perf_pt/perf_pt.c:222 [poll_loop]: aux_tail=  0x0000000000
src/backends/perf_pt/perf_pt.c:223 [poll_loop]: aux_offset=0x0000266240
src/backends/perf_pt/perf_pt.c:224 [poll_loop]: aux_size=  0x0004194304
```

Here we read 944 bytes out of the AUX buffer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # traceme
 
-A small Rust/C library to trace an arbitrary PID using Intel Processor Trace
+A small Rust/C library to trace an arbitrary thread using Intel Processor Trace
 (PT).
 
 **This is experimental code.**
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -35,12 +35,13 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use libc::{pid_t, c_char, c_void, getpid, size_t, c_int, geteuid};
+use libc::{pid_t, c_char, c_void, size_t, c_int, geteuid};
 use std::ffi::CString;
 use errors::TraceMeError;
 use std::fs::File;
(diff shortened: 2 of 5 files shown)


ltratt:
Ah, I think I was confused by the PID thing earlier. Now I know why ;) I agree, reusing `pid_t` is terrible design.

Merge as-is?

vext01:
I believe so. Go ahead.

Issue #4, opened 2017-11-28
Check for packet loss


vext01:
We should start inspecting the data buf for packet loss events.

vext01:
We can know about this by using `pt_blk_event()` and friends.

PR #26, opened 2018-01-23
Trace to memory

vext01:
This change allows us to trace to memory buffers, instead of files.

 * In doing so, we've allowed the tests to run in parallel. 
 * Once the trace pops out of C code, we say the underlying buffer is owned by `HWTrace`, hence we have a `drop` to free the buffer when it falls out of scope.
 * valgrind is happy
 * `HWTrace` will be a trait soon, as discussed.
 * A couple of semi-related bits in separate commits.

Please fire away with questions. I expect you will have loads.

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -225,11 +225,11 @@ mod tests {
     use super::PerfPTTracer;
     use ::test_helpers;
 
-    fn run_test_helper<F>(f: F)  where F: Fn(PerfPTTracer) {
+    fn run_test_helper<F>(f: F) where F: Fn(PerfPTTracer) {
         let res = PerfPTTracer::new();
         // Only run the test if the CPU supports Intel PT.
-        if res.is_ok() {
-            f(res.unwrap());
+        if let Ok(tracer) = res {
+            f(tracer);
         }
     }
 
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -69,7 +69,7 @@ pub struct PerfPTTracer {
     target_tid: pid_t,
     /// Data buffer size, in pages. Must be a power of 2.
     data_bufsize: size_t,
-    /// Aux buffer size, in pages. Must be a power of 2.
+    /// AUX buffer size, in pages. Must be a power of 2.
     aux_bufsize: size_t,
     /// Opaque C pointer representing the tracer context.
     tracer_ctx: Option<*const c_void>,
@@ -125,11 +125,11 @@ impl PerfPTTracer {
         self
     }
 
-    /// Set the PT aux buffer size.
+    /// Set the PT AUX buffer size.
     ///
     /// # Arguments
     ///
-    /// * `size` - The aux buffer size to use.
+    /// * `size` - The AUX buffer size to use.
     pub fn aux_bufsize(mut self, size: usize) -> Self {
         self.aux_bufsize = size;
         self
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -35,13 +35,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use libc::{pid_t, c_char, c_void, size_t, c_int, geteuid};
-use std::ffi::CString;
+use libc::{pid_t, c_void, size_t, c_int, geteuid};
 use errors::TraceMeError;
 use std::fs::File;
 use std::io::Read;
 use Tracer;
+use HWTrace;
 use util::linux_gettid;
+use std::ptr;
 
 // The sysfs path used to set perf permissions.
 const PERF_PERMS_PATH: &str = "/proc/sys/kernel/perf_event_paranoid";
@@ -49,7 +50,7 @@ const PERF_PERMS_PATH: &str = "/proc/sys/kernel/perf_event_paranoid";
 // FFI prototypes.
 extern "C" {
     fn perf_pt_start_tracer(conf: *const PerfPTConf) -> *const c_void;
-    fn perf_pt_stop_tracer(tr_ctx: *const c_void) -> c_int;
+    fn perf_pt_stop_tracer(tr_ctx: *const c_void, buf: *const *const u8, len: &u64) -> c_int;
 }
 
 // Struct used to communicate a tracing configuration to the C code. Must
@@ -57,7 +58,6 @@ extern "C" {
 #[repr(C)]
 struct PerfPTConf {
     target_tid: pid_t,
-    trace_filename: *const c_char,
     data_bufsize: size_t,
     aux_bufsize: size_t,
 }
@@ -65,8 +65,6 @@ struct PerfPTConf {
 /// A tracer that uses the Linux Perf interface to Intel Processor Trace.
 #[derive(Default)]
 pub struct PerfPTTracer {
-    /// Filename to store the trace to.
-    trace_filename: String,
     /// Thread ID to trace.
     target_tid: pid_t,
     /// Data buffer size, in pages. Must be a power of 2.
@@ -89,7 +87,7 @@ impl PerfPTTracer {
     ///
     /// let res = PerfPTTracer::new();
     /// if res.is_ok() {
-    ///     let tracer = res.unwrap().trace_filename("mytrace.ptt").data_bufsize(1024).target_tid(666);
+    ///     let tracer = res.unwrap().data_bufsize(1024).target_tid(666);
     /// } else {
     ///     // CPU doesn't support Intel Processor Trace.
     /// }
@@ -97,7 +95,6 @@ impl PerfPTTracer {
     pub fn new() -> Result<Self, TraceMeError> {
         if Self::pt_supported() {
             return Ok(Self {
-                trace_filename: String::from("traceme.ptt"),
                 target_tid: linux_gettid(),
                 tracer_ctx: None,
                 data_bufsize: 64,
@@ -105,16 +102,6 @@ impl PerfPTTracer {
             });
         }
         Err(TraceMeError::HardwareSupport("Intel PT not supported by CPU".into()))
-    }
-
-    /// Set where to write the trace.
-    ///
-    /// # Arguments
-    ///
-    /// * `filename` - The filename in which to store trace packets.
-    pub fn trace_filename(mut self, filename: &str) -> Self {
-        self.trace_filename = String::from(filename);
-        self
     }
 
     /// Select which thread to trace.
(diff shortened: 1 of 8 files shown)


vext01:
Sorry, spotted a small bug. Fixed.

src/backends/perf_pt/mod.rs
```diff
@@ -223,15 +205,22 @@ impl Tracer for PerfPTTracer {
...
+        let len = 0;
         let rc = unsafe {
-            perf_pt_stop_tracer(tr_ctx)
+            perf_pt_stop_tracer(tr_ctx, &buf, &len)
         };
         if rc == -1 {
             return Err(TraceMeError::CFailure);
         }
-        Ok(())
+        assert!(buf != ptr::null()); // The C code should have put the buffer address here.
```
ltratt: I'm not sure this `assert` is useful. Does the API says that it might not set `buf`? If not, we should just assume that things work (if the ptr is null, things will go badly wrong soon enough) without getting overly-defensive.

src/backends/perf_pt/mod.rs
```diff
@@ -223,15 +205,22 @@ impl Tracer for PerfPTTracer {
...
+        let trace = HWTrace::from_buf(buf, len);
+        if env::var("HWTRACE_DEBUG_DUMP_TRACES").is_ok() {
```
ltratt: This is quite an expensive call. Do we really need / want this functionality? I'm really not keen on the env vars trick: I'd prefer to have an extra function parameters or similar to control this sort of thing.

src/backends/perf_pt/mod.rs
vext01: ok

src/backends/perf_pt/perf_pt.c
```diff
@@ -105,81 +105,66 @@ struct tracer_thread_args {
...
-            return false;
-        }
+        new_data_size = head - tail;
     } else {
         // Wrap-around.
+        new_data_size = (size - tail) + head;
+    }
+
+    // For now we simply crash out if we exhaust the buffer.
+    DEBUG("current trace length: %lld", *trace_len);
```
ltratt: Is this debugging useful still? If not, I'm keen on having less output when possible.

src/backends/perf_pt/perf_pt.c
```diff
@@ -105,81 +105,66 @@ struct tracer_thread_args {
...
+    DEBUG("new_data_size: %lld", new_data_size);
+    DEBUG("trace_bufsize: %lld", trace_bufsize);
+    assert(*trace_len + new_data_size <= trace_bufsize);
+
+    // Finally copy the AUX buffer into the trace buffer.
+    if (tail <= head) {
```
ltratt: Can `tail == head`?

src/backends/perf_pt/perf_pt.c
```diff
@@ -434,31 +423,32 @@ perf_pt_start_tracer(struct tracer_conf *tr_conf)
     tr_ctx->stop_fd_rd = stop_fds[0];
     tr_ctx->stop_fd_wr = stop_fds[1];
 
-    // Use a semaphore to wait for the tracer to be ready.
-    sem_t tracer_init_sem;
-    if (sem_init(&tracer_init_sem, 0, 0) == -1) {
+    // Allocate buffer to copy the trace into (currently the same size as the AUX buffer).
```
ltratt: Not sure that "AUX" should be capitalised (it's not an acronym).

src/backends/perf_pt/mod.rs
vext01: I'm open to new ideas, but even if you have extra args, what triggers the debugging code? There needs to be an external interface.

src/lib.rs
```diff
@@ -40,12 +40,61 @@
...
+    /// Once an instance is constructed, underlying allocation and it's freeing is the
+    /// responsibility of the instance.
+    fn from_buf(buf: *const u8, len: u64) -> Self {
+        Self {buf: buf, len: len}
+    }
+
+    /// Write the trace buffer to file.
+    ///
+    /// For debugging purposes only.
+    fn to_file(&self) {
```
ltratt: This could be guarded with `#[debug]` or similar then?

src/backends/perf_pt/perf_pt.c
vext01: It's useful for development yes, but we can define `DEBUG` empty for normal usage.

src/backends/perf_pt/perf_pt.c
vext01: yes, but it's ambiguous. Either the buffer is empty or full. This is related to #4.

src/backends/perf_pt/mod.rs
ltratt: The interpreter author, IMHO, should be able to turn this on and off (if they really want to; although I can't imagine any reason for a normal interpreter author to do this). My problem with env vars as "global triggers" is that they're hard to discover, and users can occasionally trigger things that they never intended -- and may not even have realised that they've triggered something.

src/backends/perf_pt/perf_pt.c
vext01: Until we implement that, it's assumed to be empty.

src/backends/perf_pt/perf_pt.c
vext01: can do.

src/lib.rs
vext01: sure, didn't know about that!

src/backends/perf_pt/perf_pt.c
ltratt: What I'm saying is that there's always a temptation to leave lots of print statements (which this is) lurking around that are no longer useful. I'm very keen not to leave anything around that isn't still actively useful. If we're not careful we'll end up like one of those programs that spews stuff to the terminal forever, because no-one's brave enough to remove the print statements.

src/backends/perf_pt/perf_pt.c
ltratt: You've lost me.

src/backends/perf_pt/perf_pt.c
vext01: If we define `DEBUG` empty, nothing is printed. I would like to keep the DEBUG statements in.

ltratt: We have to be really careful. 95% of debugging statements left in programs are not only not useful, but they actively make finding real problems harder. Here's one guide I sometimes use: would a normal user ever want to see this output? Could it be useful to them? If not, we should remove them. Remember, adding print statements in for temporary debugging is super cheap.

vext01: > would a normal user ever want to see this output? 

No, but they wouldn't anyway. It's like `DNPRINTF` in the OpenBSD kernel. You turn it on at compile time if you want it.

If you really feel very strongly, we can kill it, but I'm telling you, I do find this code very useful.

src/backends/perf_pt/perf_pt.c
vext01: The aux buffer is a ringer buffer.

It starts with `head == tail`, and as data goes in, `head` zips along the buffer. If you don't take data out quick enough, it eventually wraps around. PT provides a way to know if this happens, but we have not implemented it yet.

vext01: [if it happens, then we lost packets]

src/backends/perf_pt/perf_pt.c
ltratt: At the very least, let's think about reducing the output. For example, the two lines that come after this one, seem superfluous because the `assert` catches any violations, and they seem to be the result of defensive programming. I'm not sure the "read with wrap/no wrap" is useful either, but maybe I'm wrong.

src/backends/perf_pt/mod.rs
vext01: OK. Well I guess that moves the problem up to the interpreter at least. They'd still need switches or environment vars, right?

I'm OK with that if you are.

ltratt: Yes, they'd be in charge of turning this on or off (however they want to).

src/backends/perf_pt/perf_pt.c
vext01: That sounds more reasonable.

I think the main goal of these `DEBUG`s is to know roughly where to start looking without having to do a binary chop to find out.

Perhaps I could try to limit the statements to one at the top of important-ish functions? Would you be OK with that?

ltratt: I wouldn't like to put that sort of rule in place. I'm really not keen on these statements lurking around before we know they might be useful. They clog up code something rotten and, when done pre-emptively, tend not to produce very useful output IME. More is not always better ;)

src/backends/perf_pt/mod.rs
vext01: The only case where ptr is set to null, is in failure, which should be caught by the previous line.

ltratt: OK, we can kill this line then.

src/backends/perf_pt/perf_pt.c
vext01: On second thoughts, the manual page calls it `AUX`.

E.g.

> The AUX region allows mmaping a separate sample buffer for high-bandwidth data streams (separate from the main perf sample buffer).

src/lib.rs
vext01: I think an earlier discussion invalidates this now. The VM author needs to be able to call this.

src/backends/perf_pt/perf_pt.c
ltratt: Fair enough. I vaguely remember somewhere where it wasn't capitalised thus, but I might be wrong. Maybe quickly check?

src/lib.rs
ltratt: Two dumb questions. Why would they need to do this? And what's to stop them writing a function to do it themselves?

src/backends/perf_pt/perf_pt.c
vext01: I've killed all of the calls to `DEBUG`, but left the macro in, since I don't want to have to write that afresh every time.

vext01:
I've address your comments Laurie.

Squashing required before merge.

src/backends/perf_pt/perf_pt.c
vext01: Fixed. Thanks.

ltratt:
This mostly looks OK, but I'm still not convinced we need the "write to file" function. I'm prepare to be convinced though!

vext01:
In this repo:
https://github.com/01org/processor-trace

There's a load of utilities for inspecting on-disk traces at different levels. I've been using them during development, so I'd like a way to get a trace onto disk without too much effort.

ltratt:
I know it's a nice-to-have in some ways, but we also need to bear in mind the cognitive cost of every extra function in a public API. Is it worth it if there's another way that people can access the memory and write their own write-to-file function.

vext01:
I think we can kill it eventually, but until we can decode the instruction stream, removing it will only slow me down.

If I could do something like `#[debug]` I would, but it seems this doesn't exist.

ltratt:
Is it something like `#[cfg(debug)]`?

vext01:
Obviously it's `#[cfg(debug_assertions)]` :P

Fix pushed.

vext01:
OK, that's about as good as it gets with the `use` logic. Some still need to be outside the func since they appear in the function sig.

ltratt:
If you squash, I can merge.

vext01:
Squashed. Since there were quite a few little "unrelated" bits, I've left those in separate commits.

PR #31, opened 2018-01-26
Make the IPT tracer re-usable and document the mmap(2) dance.

vext01:
This change makes the tracers efficiently re-usable.

Get some popcorn, because there's a bit of back-story before you can review.

In short, before we would `perf_event_open()` a new perf fd and `mmap()` our buffers every time we started tracing. This is inefficient as these can be re-used over and over. This change fixes that.

However, it comes with an API change which I didn't appreciate until implementing this. Before we had:
```
let tracer = PerfPTTracer::new().target_tid(12345).aux_bufsize(64);
tracer.start();
... work
let trace = tracer.stop()
```

Now that the tracer is re-usable, under this interface, the user might think they can re-configure the tracer before subsequent calls to `start_tracing()`:

```
let tracer = PerfPTTracer::new().target_tid(12345).aux_bufsize(64);
tracer.start();
... work
let trace1 = tracer.stop()

let tracer = tracer.target_tid(33333); // naughty! bad user!

tracer.start();
... work
let trace2 = tracer.stop()
```

This won't work because the file descriptor stored inside the `PerfPTTracer` is specific to a thread id (tid). AFAICS, the `perf_event_open()` API doesn't allow the tid to be changed after the fact. That's fine, and it actually fits nicely with our intention -- we don't want to share a these tracers between threads. We will have one per thread.

Although the buffer sizes *could* be changed on an existing tracer, I don't see any good reason to allow this, and it adds complexity.

My proposal is to have an API which "locks in" the configuration when the tracer instance is created:
```
let config = PerfPTTracer::config(); // gets the default config.
let config = config.target_tid(12345).aux_bufsize(64);
let tracer = PerfPTTracer::new(config); // config is now gone forever.
tracer.start();
... work
let trace1 = tracer.stop()

// There is nothing the user can do which might imply the configuration can be changed here.

tracer.start();
... work
let trace2 = tracer.stop()
```

Other notes:
 * On the Rust side, we can't use the existence of the opaque pointer to know if we are tracing or not, so I've added a Boolean flag.

 * On the C side, in order to split the initialisation from the starting of the tracer, I've had to understand the `mmap(2)` dance used to expose the kernel buffers. This code is fiddly and trips me up every time I look at it. To help my future self I've simplified the code a bit (and managed to eliminate a fair amount of sharing between threads) and added some comments. [In hindsight I would have liked to have that as a separate commit, but it will be hard to pick apart now, sorry.]

 * Checked `ptxed` can decode the trace when dumped to disk. All is well.

 * Tested under valgrind, and with error states too. All allocations freed, and fds closed.

 * All tests, including a new test pass.

 * Example amended to show repeated tracing.

Comments? OK?

### DIFF
--- a/src/backends/dummy.rs
+++ b/src/backends/dummy.rs
@@ -39,34 +39,61 @@ use Tracer;
 use errors::HWTracerError;
 use HWTrace;
 use std::ptr;
+#[cfg(debug_assertions)]
+use std::ops::Drop;
+use TracerState;
 
 /// A tracer which doesn't really do anything.
 pub struct DummyTracer {
-    started: bool,
+    state: TracerState,
 }
 
 impl DummyTracer {
     /// Create a dummy tracer.
     pub fn new() -> Self {
-        Self {started: false}
+        Self { state: TracerState::Stopped }
+    }
+
+    fn err_if_destroyed(&self) -> Result<(), HWTracerError> {
+        if self.state == TracerState::Destroyed {
+            return Err(HWTracerError::TracerDestroyed);
+        }
+        Ok(())
     }
 }
 
 impl Tracer for DummyTracer {
     fn start_tracing(&mut self) -> Result<(), HWTracerError> {
-        if self.started {
+        self.err_if_destroyed()?;
+        if self.state != TracerState::Stopped {
             return Err(HWTracerError::TracerAlreadyStarted);
         }
-        self.started = true;
+        self.state = TracerState::Started;
         Ok(())
     }
 
     fn stop_tracing(&mut self) -> Result<HWTrace, HWTracerError> {
-        if !self.started {
+        self.err_if_destroyed()?;
+        if self.state != TracerState::Started {
             return Err(HWTracerError::TracerNotStarted);
         }
-        self.started = false;
+        self.state = TracerState::Stopped;
         Ok(HWTrace::from_buf(ptr::null(), 0)) // An empty trace.
+    }
+
+    fn destroy(&mut self) -> Result<(), HWTracerError> {
+        self.err_if_destroyed()?;
+        self.state = TracerState::Destroyed;
+        Ok(())
+    }
+}
+
+#[cfg(debug_assertions)]
+impl Drop for DummyTracer {
+    fn drop(&mut self) {
+        if self.state != TracerState::Destroyed {
+            panic!("DummyTracer dropped without destroy()");
+        }
     }
 }
 
@@ -81,6 +108,11 @@ mod tests {
     }
 
     #[test]
+    fn test_repeated_tracing() {
+        test_helpers::test_repeated_tracing(DummyTracer::new());
+    }
+
+    #[test]
     fn test_already_started() {
         test_helpers::test_already_started(DummyTracer::new());
     }
@@ -88,5 +120,27 @@ mod tests {
     #[test]
     fn test_not_started() {
         test_helpers::test_not_started(DummyTracer::new());
+    }
+
+    #[test]
+    fn test_use_tracer_after_destroy1() {
+        test_helpers::test_use_tracer_after_destroy1(DummyTracer::new());
+    }
+
+    #[test]
+    fn test_use_tracer_after_destroy2() {
+        test_helpers::test_use_tracer_after_destroy2(DummyTracer::new());
+    }
+
+    #[test]
+    fn test_use_tracer_after_destroy3() {
+        test_helpers::test_use_tracer_after_destroy3(DummyTracer::new());
+    }
+
+    #[cfg(debug_assertions)]
+    #[should_panic]
+    #[test]
+    fn test_drop_without_destroy() {
+        test_helpers::test_drop_without_destroy(DummyTracer::new());
     }
 }
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -43,65 +43,42 @@ use Tracer;
 use HWTrace;
 use util::linux_gettid;
 use std::ptr;
+#[cfg(debug_assertions)]
+use std::ops::Drop;
+use TracerState;
 
 // The sysfs path used to set perf permissions.
 const PERF_PERMS_PATH: &str = "/proc/sys/kernel/perf_event_paranoid";
 
 // FFI prototypes.
 extern "C" {
-    fn perf_pt_start_tracer(conf: *const PerfPTConf) -> *const c_void;
+    fn perf_pt_init_tracer(conf: *const PerfPTConf) -> *const c_void;
+    fn perf_pt_start_tracer(tr_ctx: *const c_void) -> c_int;
     fn perf_pt_stop_tracer(tr_ctx: *const c_void, buf: *const *const u8, len: &u64) -> c_int;
+    fn perf_pt_free_tracer(tr_ctx: *const c_void) -> c_int;
 }
 
 // Struct used to communicate a tracing configuration to the C code. Must
 // stay in sync with the C code.
 #[repr(C)]
-struct PerfPTConf {
-    target_tid: pid_t,
-    data_bufsize: size_t,
-    aux_bufsize: size_t,
-}
-
-/// A tracer that uses the Linux Perf interface to Intel Processor Trace.
-#[derive(Default)]
-pub struct PerfPTTracer {
+pub struct PerfPTConf {
     /// Thread ID to trace.
     target_tid: pid_t,
     /// Data buffer size, in pages. Must be a power of 2.
     data_bufsize: size_t,
     /// AUX buffer size, in pages. Must be a power of 2.
     aux_bufsize: size_t,
-    /// Opaque C pointer representing the tracer context.
-    tracer_ctx: Option<*const c_void>,
 }
 
-impl PerfPTTracer {
-    /// Create a new tracer.
-    ///
-    /// Returns `Err` if the CPU doesn't support Intel Processor Trace.
-    ///
-    /// # Example
-    ///
-    /// ```
-    /// use hwtracer::backends::PerfPTTracer;
-    ///
-    /// let res = PerfPTTracer::new();
-    /// if res.is_ok() {
-    ///     let tracer = res.unwrap().data_bufsize(1024).target_tid(666);
-    /// } else {
-    ///     // CPU doesn't support Intel Processor Trace.
-    /// }
-    /// ```
-    pub fn new() -> Result<Self, HWTracerError> {
-        if Self::pt_supported() {
-            return Ok(Self {
-                target_tid: linux_gettid(),
-                tracer_ctx: None,
-                data_bufsize: 64,
-                aux_bufsize: 1024,
-            });
+/// Configures a PerfPTTracer.
+impl PerfPTConf {
+    /// Creates a new configuration with defaults.
+    pub fn new() -> Self {
+        Self {
+            target_tid: linux_gettid(),
+            data_bufsize: 64,
+            aux_bufsize: 1024,
         }
-        Err(HWTracerError::HardwareSupport("Intel PT not supported by CPU".into()))
     }
 
     /// Select which thread to trace.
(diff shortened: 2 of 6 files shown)


src/backends/perf_pt/mod.rs
```diff
@@ -179,39 +197,40 @@ impl PerfPTTracer {
     }
 }
 
+impl Drop for PerfPTTracer {
+    fn drop(&mut self) {
+        let res = unsafe { perf_pt_free_tracer(self.tracer_ctx) };
```
ltratt: This worries me a bit. It looks to me like `perf_pt_free_tracer` is the sort of call which will fail at some point... and then... boom, the whole program dies. Can we / should we make users manually free things?

src/backends/perf_pt/perf_pt.c
```diff
@@ -348,124 +318,173 @@ tracer_thread(void *arg)
...
+
+    // Allocate mmap(2) buffers for speaking to perf.
+    //
+    // These regions are mmap(2)ed from the perf fd in two separate chunks:
+    //
+    //            "Base buffer"
+    // -----------------------------------     ------------------
+    // | header  |       data buffer     | ... |   AUX buffer   |
+    // -----------------------------------     ------------------
+    //           ^ header->data_offset
```
ltratt: Is the AUX buffer immediately after the data buffer? If so, how does one know where the data buffer stops and the AUX buffer starts?

src/backends/perf_pt/perf_pt.c
```diff
@@ -348,124 +318,173 @@ tracer_thread(void *arg)
...
+    //
+    // The AUX buffer is where the kernel exposes control flow packets, whereas
+    // the data buffer is used for all other kinds of packet.
+
+    // Allocate the base buffer.
+    //
+    // Data buffer is preceded by one management page (the header), hence `1 +
```
ltratt: The header mst be exactly 1 page big? What happens if the page size changes?

src/backends/perf_pt/perf_pt.c
```diff
@@ -348,124 +318,173 @@ tracer_thread(void *arg)
...
         goto clean;
     }
 
-    ret = tr_ctx;
 clean:
-    if (failing) {
-        ret = NULL;
-        // Child thread must be stopped before closing the shared perf_fd.
+    if (clean_sem) {
+        if (sem_destroy(&tracer_init_sem) == -1) {
```
ltratt: Can be folded with `&&` into the outer `if`.

src/backends/perf_pt/perf_pt.c
```diff
@@ -503,25 +523,64 @@ perf_pt_stop_tracer(struct tracer_ctx *tr_ctx, uint8_t **buf, __u64 *len) {
...
+/*
+ * Clean up and free a tracer_ctx and its contents.
+ *
+ * Returns 0 on success, or -1 otherwise.
+ */
+int
+perf_pt_free_tracer(struct tracer_ctx *tr_ctx) {
+    int ret = 0;
+
+    if (tr_ctx->aux_buf) {
```
ltratt: Can be simplified with `&&` (similarly below).

src/backends/perf_pt/mod.rs
vext01: We could, but it's not very Rusty, is it?

src/backends/perf_pt/perf_pt.c
vext01: They are at consecutive offsets in the file descriptor, but get mmaped to different virtual address ranges (we don't provide an mmap hint).

src/backends/perf_pt/perf_pt.c
vext01: Whilst the system is running? I think you would have other issues ;)

src/backends/perf_pt/perf_pt.c
vext01: True, will fix.

src/backends/perf_pt/mod.rs
vext01: [FWIW, I don't expect this to ever fail]

src/backends/perf_pt/perf_pt.c
vext01: > how does one know where the data buffer stops and the AUX buffer starts?

You hold two different pointers:
```
tr_ctx->base_buf = mmap(...
...
tr_ctx->aux_buf = mmap(...
```

src/backends/perf_pt/mod.rs
ltratt: I don't agree: I think it *will* fail at some point. Every Unix API that can lead to failure does, I've learnt, lead to failure at some point in practise. I know that `Drop` is easy to use, but I think we'll end up having some sort of manual way of doing things anyway in the long run, so we might as well bite the bullet now.

src/backends/perf_pt/perf_pt.c
ltratt: I mean: it's a weird way of specifying a data-structure size. I'm checking that I understood it correctly!

src/backends/perf_pt/perf_pt.c
ltratt: OK, I think you need to update the explanation a bit. This isn't very clear at the moment: the figure suggests the two buffers follow each other (somehow).

vext01: Well, we don't know exactly where there are relative to each other, other than, they are non-overlapping. Anywhere I put the box, there is another place the box could also be :P.

Is it the `...` between the boxes that is confusing?

ltratt: Maybe just don't put the AUX buffer in the diagram then? I don't think it helps.

src/backends/perf_pt/mod.rs
vext01: Why would it fail? If it did, it would certainly be a bug on our part.

This will only fail if `munmap`, `close`, `free`, or `pthread_join` fails.

If one of those were to fail, what would the user do with the error, which basically says "something really weird happened"?

ltratt: `EINTR` if nothing else...

src/backends/perf_pt/perf_pt.c
vext01: How about a more textual description?

ltratt: Worth a shot.

src/backends/perf_pt/perf_pt.c
vext01: It is weird, I agree.

But the page size is fixed at kernel compile time, so don't worry about that.

FWIW, here's the same in the libipt docs:

ltratt: Missing text?

vext01: yeah, hit the wrong button, updated.

src/backends/perf_pt/mod.rs
vext01: It's true that close can return `EINTR` and it's never been clear what you should do if it does:
https://lwn.net/Articles/576478/

We can ask the user to explicitly check a result from a manual free, but at which point all bets are kind of off.

So what do you think we should we do? Still make it explicit?

src/backends/perf_pt/perf_pt.c
ltratt: OK.

src/backends/perf_pt/mod.rs
ltratt: There are other ways for things to go wrong, but this is one. I've seen `close` fail too. So, yes, I really fear we have to make this explicit :/

ltratt: I meant `pthread_join` -- sorry!

vext01: OK, let's do that. Although I think it's a shame that we couldn't make it more Rusty.

ltratt: This is the same sort of problem ones have with GC: finalisers look really useful, until you realise that at times like this, they're not quite the right thing.

vext01:
Try this Laurie.

It occurs to me that one thing we could do is do "something" if the user forgets to `destroy()` a tracer. We can have a `drop()` which either `panic()s` or if it's a debug build prints something to stderr?

Too far?

vext01:
Let's try and get this in today.

src/backends/perf_pt/perf_pt.c
```diff
@@ -346,14 +346,18 @@ perf_pt_init_tracer(struct tracer_conf *tr_conf)
...
-    // | header  |       data buffer     | ... |   AUX buffer   |
-    // -----------------------------------     ------------------
+    // 1) The base buffer (tr_ctx->base_buf), which looks like this:
+    //
+    // -----------------------------------
+    // | header  |       data buffer     |
+    // -----------------------------------
     //           ^ header->data_offset
     //
+    // 2) The AUX buffer (tr_ctx->aux_buf), which is a simple array of bytes.
```
ltratt: I don't think "which is a simple array of bytes" adds anything useful, and can be deleted.

src/backends/perf_pt/perf_pt.c
```diff
@@ -467,10 +471,8 @@ perf_pt_start_tracer(struct tracer_ctx *tr_ctx)
     }
 
 clean:
-    if (clean_sem) {
-        if (sem_destroy(&tracer_init_sem) == -1) {
-            ret = -1;
-        }
+    if ((clean_sem) && (sem_destroy(&tracer_init_sem) == -1)) {
```
ltratt: `(clean_sem)` doesn't need to be bracketed.

src/backends/perf_pt/perf_pt.c
```diff
@@ -552,15 +554,13 @@ int
...
-        }
+    if ((tr_ctx->aux_buf) &&
+        (munmap(tr_ctx->aux_buf, tr_ctx->aux_bufsize) == -1)) {
+        ret = -1;
     }
-    if (tr_ctx->base_buf) {
-        if (munmap(tr_ctx->base_buf, tr_ctx->base_bufsize) == -1) {
-            ret = -1;
-        }
+    if ((tr_ctx->base_buf) &&
```
ltratt: `(tr_ctx->base_buf)` doesn't need to be bracketed.

src/backends/dummy.rs
```diff
@@ -43,17 +43,29 @@ use std::ptr;
 /// A tracer which doesn't really do anything.
 pub struct DummyTracer {
     started: bool,
+    destroyed: bool,
```
ltratt: These two bools look like they're forming a state machine to me. Maybe there should be a single `enum` which stores the state?

ltratt:
Sorry, I missed this one :/ I quite like the idea of `Drop` being an `assert` (probably a `debug_assert`) that checks that the trace has been manually destroyed elsewhere.

vext01:
ok. I'll get onto that eventually.

src/backends/dummy.rs
vext01: good idea.

vext01:
That's those two fixes implemented. Looks good?

src/backends/perf_pt/mod.rs
```diff
@@ -275,7 +286,7 @@ mod tests {
 
     #[test]
     fn test_repeated_tracing() {
-        run_test_helper(test_helpers::test_repeated_tracing);
+            run_test_helper(test_helpers::test_repeated_tracing);
```
vext01: Grrr.

ltratt:
Dumb question: should the state struct be duplicated between tracers or shared? I think the shared API implicitly assumes stopped/started/destroyed, so maybe it should a common struct that the various tracers use?

vext01:
I did ask myself the same question. Shall we try sharing it, and then we can think again if something explodes later?

ltratt:
Works for me.

vext01:
Implemented. Ready for squashing?

ltratt:
Looks good to me. Please squash away!

PR #32, opened 2018-02-02
Use a trait for traces, so that different backends can return different traces.

vext01:
This change (as discussed earlier somewhere) uses a generic type for traces coming from different backends.

Looks good?

### DIFF
--- a/src/backends/dummy.rs
+++ b/src/backends/dummy.rs
@@ -37,11 +37,18 @@
 
 use Tracer;
 use errors::HWTracerError;
-use HWTrace;
-use std::ptr;
+use Trace;
 #[cfg(debug_assertions)]
 use std::ops::Drop;
 use TracerState;
+
+/// An empty dummy trace.
+struct DummyTrace {}
+
+impl Trace for DummyTrace {
+    #[cfg(debug_assertions)]
+    fn to_file(&self, _: &str) {}
+}
 
 /// A tracer which doesn't really do anything.
 pub struct DummyTracer {
@@ -72,13 +79,13 @@ impl Tracer for DummyTracer {
         Ok(())
     }
 
-    fn stop_tracing(&mut self) -> Result<HWTrace, HWTracerError> {
+    fn stop_tracing(&mut self) -> Result<Box<Trace>, HWTracerError> {
         self.err_if_destroyed()?;
         if self.state != TracerState::Started {
             return Err(HWTracerError::TracerNotStarted);
         }
         self.state = TracerState::Stopped;
-        Ok(HWTrace::from_buf(ptr::null(), 0)) // An empty trace.
+        Ok(Box::new(DummyTrace{}))
     }
 
     fn destroy(&mut self) -> Result<(), HWTracerError> {
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -35,17 +35,16 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use libc::{pid_t, c_void, size_t, c_int, geteuid};
+use libc::{pid_t, c_void, size_t, c_int, geteuid, free};
 use errors::HWTracerError;
 use std::fs::File;
 use std::io::Read;
 use Tracer;
-use HWTrace;
 use util::linux_gettid;
 use std::ptr;
 #[cfg(debug_assertions)]
 use std::ops::Drop;
-use TracerState;
+use {TracerState, Trace};
 
 // The sysfs path used to set perf permissions.
 const PERF_PERMS_PATH: &str = "/proc/sys/kernel/perf_event_paranoid";
@@ -57,6 +56,54 @@ extern "C" {
     fn perf_pt_stop_tracer(tr_ctx: *const c_void, buf: *const *const u8, len: &u64) -> c_int;
     fn perf_pt_free_tracer(tr_ctx: *const c_void) -> c_int;
 }
+
+/// A raw Intel PT trace, obtained via Linux perf.
+pub struct PerfPTTrace {
+    buf: *const u8,
+    #[allow(dead_code)]
+    len: u64,
+}
+
+impl PerfPTTrace {
+    /// Makes a new trace from a raw pointer and a size.
+    ///
+    /// The `buf` argument is assumed to have been allocated on the heap using malloc(3). `len`
+    /// must be less than or equal to the allocated size.
+    ///
+    /// The allocation is automatically freed by Rust when the struct falls out of scope.
+    fn from_buf(buf: *const u8, len: u64) -> Self {
+        Self {buf: buf, len: len}
+    }
+}
+
+impl Trace for PerfPTTrace {
+    /// Write the raw trace packets into the specified file.
+    ///
+    /// This can be useful for developers who want to use (e.g.) the pt utility to inspect the raw
+    /// packet stream.
+    #[cfg(debug_assertions)]
+    fn to_file(&self, filename: &str) {
+        use std::slice;
+        use std::fs::File;
+        use std::io::prelude::*;
+
+        let mut f = File::create(filename).unwrap();
+        let slice = unsafe { slice::from_raw_parts(self.buf, self.len as usize) };
+        f.write(slice).unwrap();
+    }
+}
+
+
+/// Once a PerfPTTrace is brought into existence, we say the instance owns the C-level allocation.
+/// When the it falls out of scope, free up the memory.
+impl Drop for PerfPTTrace {
+    fn drop(&mut self) {
+        if self.buf != ptr::null() {
+            unsafe { free(self.buf as *mut c_void) };
+        }
+    }
+}
+
 
 // Struct used to communicate a tracing configuration to the C code. Must
 // stay in sync with the C code.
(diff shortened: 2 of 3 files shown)


src/backends/dummy.rs
```diff
@@ -37,12 +37,19 @@
...
 use Tracer;
 use errors::HWTracerError;
-use HWTrace;
-use std::ptr;
+use Trace;
 #[cfg(debug_assertions)]
 use std::ops::Drop;
 use TracerState;
 
+/// An imaginary trace.
```
ltratt: I don't think this comment is entirely helpful: it's really not an imaginary trace. It's a trace that can only ever be empty, right?

src/backends/perf_pt/mod.rs
```diff
@@ -58,6 +57,55 @@ extern "C" {
     fn perf_pt_free_tracer(tr_ctx: *const c_void) -> c_int;
 }
 
+/// A Intel PT trace, obtained via Linux Perf.
```
ltratt: "A*n* Intel". Also is "perf" capitalised or not (I'm not sure)?

src/backends/perf_pt/mod.rs
```diff
@@ -58,6 +57,55 @@ extern "C" {
...
+#[allow(dead_code)]
```
ltratt: Why do we allow dead code here? I'm more curious than anything!

src/backends/dummy.rs
vext01: Yes, an entirely vacuous container, only existing to satisfy a type constraint really.

src/backends/perf_pt/mod.rs
```diff
@@ -58,6 +57,55 @@ extern "C" {
...
+pub struct PerfPTTrace {
+    buf: *const u8,
+    len: u64,
+}
+
+impl PerfPTTrace {
+    /// Makes a new trace from a raw pointer and a size.
+    ///
+    /// The `buf` argument is assumed to have been allocated on the heap using malloc(3). `len`
```
ltratt: I don't like the fact that the user calls `malloc` but we call `free`. Why don't we just take `len` as an arg and then `malloc` buf ourselves?

src/backends/perf_pt/mod.rs
vext01: Yeah, you are right. Lower case in the kernel docs.

src/backends/perf_pt/mod.rs
```diff
@@ -58,6 +57,55 @@ extern "C" {
...
+        f.write(slice).unwrap();
+    }
+}
+
+
+/// Once a PerfPTTrace is brought into existence, we say the instance owns the C-level allocation.
+/// When the it falls out of scope, free up the memory.
+impl Drop for PerfPTTrace {
+    fn drop(&mut self) {
+        if self.buf != ptr::null() {
```
ltratt: How could `self.buf` be `null`?

src/backends/dummy.rs
ltratt: Them's some big words! How about just "A trace which can only ever be empty" or similar? My brain does better with short words ;)

src/backends/perf_pt/mod.rs
vext01: The user doesn't call malloc, the underlying C code does.

Why are we telling the user about this, is my reflexive question now?! This should probably be mentioned in a non-doc comment somewhere really. Will fix.

ltratt: The doc comment says that the user must have malloc'd the memory that `buf` points to; and we later call `free(buf)` ourselves. Experience suggests that's not a pleasant API. Except in really weird circumstances (which I don't think we have here), it's better to have either the user do both `malloc` & `free` or us do both `malloc` and `free`.

src/backends/perf_pt/mod.rs
vext01: If `start_tracing` or `stop_tracing` failed for example.

https://github.com/softdevteam/hwtracer/blob/9902d858002e5bd3a4e599f97ea8c7f3a7285122/src/backends/perf_pt/perf_pt.c#L488

https://github.com/softdevteam/hwtracer/blob/9902d858002e5bd3a4e599f97ea8c7f3a7285122/src/backends/perf_pt/perf_pt.c#L542

src/backends/perf_pt/mod.rs
vext01: The user does neither.

src/backends/perf_pt/mod.rs
vext01: In release mode, we get:
```
warning: field is never used: `len`
  --> src/backends/perf_pt/mod.rs:63:5
   |
63 |     len: u64,
   |     ^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default
```

src/backends/perf_pt/mod.rs
ltratt: OK, let's say "user of this API". I know it's an internal user -- but it's still a user. Is there a reason we can't `malloc` the buffer from inside Rust?

src/backends/perf_pt/mod.rs
ltratt: Can we put that just on that single attribute? I don't know if it works though...

vext01: This will change once we decode the packets, of course. We could leave the annotation off so we don't forget to remove it later?

src/backends/perf_pt/mod.rs
ltratt: Maybe I'm an idiot, but at least the first of those seems to be writing to a value that isn't `self.buf`?

src/backends/perf_pt/mod.rs
ltratt: If it works just on that one attribute, I think it's fine. Or maybe put the `cfg(debug_assertions)` hack on the field instead?

src/backends/perf_pt/mod.rs
vext01: That API would be possible, yes, it would have to look like this:

```
let tracer = ...;
let mut trace = tracer::empty_trace();  // does malloc
tracer.start_tracing(&mut trace);
...
tracer.stop_tracing();
... do stuff with trace, buf freed when falls out of scope.
```

We could try this. It's not what we agreed on earlier, but then I guess we didn't think too hard about the consequences of C-level allocations.

One thing I do insist upon though, is that C code is able to `realloc(3)` the buffer. This means rust must be willing for the pointer to move.

Thoughts?

vext01: To be clear, the user of the API doesn't care currently. The buffer is allocated by `start_tracing` and freed by `drop()`. They never see the memory ops.

src/backends/perf_pt/mod.rs
vext01: `#[allow(dead_code)]` works on the one field! Neat.

src/backends/perf_pt/mod.rs
ltratt: Couldn't it be something like:

```rust
let tracer = ...;
tracer.start_tracing();
let trace = tracer.stop_tracing();
```

I guess `start_tracing` would say something like "if `buf` is null, `malloc` something" and `stop_tracing` would move the buffer to the trace. `Drop` would make sense on both the trace and tracer in this case, I think.

I admit that I'm not sure where to handle the "C code can realloc" case. Maybe a special method in the tracer `update_buf_ptr` or similar?

src/backends/perf_pt/mod.rs
ltratt: Wouldn't the `cfg(debug_assertions)` thing be better? It means if we end up deleting all code that uses the attribute, we'll get a compile-time error on the field.

src/backends/perf_pt/mod.rs
vext01: It's the same buffer.

The buffer first lives in ` tr_ctx->trace_buf` until the user stops tracing, then "ownership" is transferred to the Rust `PerfPTTrace` struct here:

https://github.com/softdevteam/hwtracer/blob/9902d858002e5bd3a4e599f97ea8c7f3a7285122/src/backends/perf_pt/perf_pt.c#L534

Where `buf` is a pointer to the `self.buf` on the Rust side.

[This relates to your other question about allocations]

src/backends/perf_pt/mod.rs
vext01: We would have to delete the code which *sets* len too. Read the error again :)

src/backends/perf_pt/mod.rs
vext01: > I guess start_tracing would say something like "if buf is null, malloc something"

That doesn't quite make sense, but I know roughly what you are getting at [I think]. We could try this but I suggest *not* in this PR.

> I admit that I'm not sure where to handle the "C code can realloc" case. Maybe a special method in the tracer update_buf_ptr or similar?

This can be solved with pointers to pointers.

vext01: To expand upon that, these lines:

https://github.com/softdevteam/hwtracer/blob/9902d858002e5bd3a4e599f97ea8c7f3a7285122/src/backends/perf_pt/perf_pt.c#L154-L155

Would be replaced with a call to `realloc` which writes to a pointer to a pointer to the allocation. Do you get it?

vext01: I've raised #33 for this.

vext01: Oh and #34.

Are you OK for these to be separate PRs? (the realloc one can wait for a while actually).

vext01:
I've addressed the small comments and raised issues for the bigger comments :)

Let me know if I can squash.

src/backends/perf_pt/mod.rs
ltratt: OK, I admit that the split between Rust and C code makes this a bit hard to understand in isolation.

src/backends/perf_pt/mod.rs
ltratt: Fine with me.

ltratt:
Yep, please squash.

vext01:
splat

Issue #33, opened 2018-02-02
Allow trace buffer to grow


vext01:
Fix this:
https://github.com/softdevteam/hwtracer/blob/9902d858002e5bd3a4e599f97ea8c7f3a7285122/src/backends/perf_pt/perf_pt.c#L154-L155

PR #35, opened 2018-02-06
Do trace storage allocation on the Rust side.

vext01:
This change ensures that Rust does both the `malloc(3)` and `free(3)` of the trace buffers.

This was hard, but I think I got there in the end. valgrind happy, all tests pass.

Looks good?

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -35,7 +35,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use libc::{pid_t, c_void, size_t, c_int, geteuid, free};
+use libc::{pid_t, c_void, size_t, c_int, geteuid, malloc, free};
 use errors::HWTracerError;
 use std::fs::File;
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


src/backends/perf_pt/mod.rs
```diff
@@ -260,10 +278,18 @@ impl Tracer for PerfPTTracer {
             return Err(HWTracerError::TracerAlreadyStarted);
         }
 
-        if unsafe { perf_pt_start_tracer(self.tracer_ctx) } == -1 {
+        // It is essential we box the trace now to stop it from moving. If it were to move, then
+        // the reference which we pass to C here would become invalid. The interface to
```
ltratt: Is this statement true? I assume you're not passing a poiunter to `PerfPTTrace`, at which point it doesn't matter if `PerfPTTrace` moves, so long as the C memory inside it doesn't?

src/backends/perf_pt/perf_pt.c
```diff
@@ -400,14 +406,20 @@ perf_pt_init_tracer(struct tracer_conf *tr_conf)
...
  * Returns zero on success, or -1 otherwise.
  */
 int
-perf_pt_start_tracer(struct tracer_ctx *tr_ctx)
+perf_pt_start_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_trace *trace)
 {
     int clean_sem = 0, clean_thread = 0;
     int ret = 0;
 
+
```
ltratt: Double blank line?

src/backends/perf_pt/mod.rs
vext01: We are passing a pointer to a `PerfPTTrace`, as C needs to access the other fields too. 

Even if we were only passing the buffer itself, we would need to pass a `*mut *mut u8` which would have the same problem. Although the underlying C-level heap allocation won't move, the intermediate pointer might. The intermediate pointer is needed for a `realloc(3)` when we get around to implementing #33.

[I actually found this problem *before* i was passing a reference to the whole struct]

src/backends/perf_pt/perf_pt.c
vext01: Oops, yes. That'll be a debugging leftover.

vext01:
Comments addressed.

src/backends/perf_pt/mod.rs
ltratt: OK, I think I get it.

ltratt:
If you squash, I can merge.

vext01:
splat

src/backends/perf_pt/mod.rs
vext01: It's a mind bender, I agree. It took me most part of a day to find this bug!

PR #38, opened 2018-02-12
Use bool return type in the C api.

### LINKED ISSUE #37: Use `bool`s in API
Rust finally made a decision about the ABI meaning of `bool`, so we can use `bool` in our foreign APIs if we like:

https://github.com/rust-lang/rfcs/issues/1982#issuecomment-364647717

vext01:
Rust recently decided that a Rust `bool` is ABI compatible with a C99
`_Bool` and thus a C99 `bool`.

Fixes #37


ltratt:
I thought it would take them much longer than this to stabilise this feature, but I'm happy to be wrong!

vext01:
Indeed!

PR #39, opened 2018-02-16
Depend on libipt.

vext01:
This is the {fetch and build, use system} libipt build system change we discussed the other day.

No functional change. In preparation for decoding IPT traces. Dummy code is included to check linkage works.

Tested with/without bundled build of libipt.

Looks good?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,9 @@ This library only supports Intel Processor Trace, but in the future we hope to
 support alternatives such as Arm's CoreSight.
 
 **This is experimental code.**
+
+## Notes
+
+When running `cargo`, you can set `IPT_PATH=...` to specify a path to a system
+libipt to use. If this variable is absent, Cargo will download and build libipt
+for you.
--- a/build.rs
+++ b/build.rs
@@ -38,10 +38,14 @@
 extern crate gcc;
 
 #[cfg(target_os = "linux")]
-use std::path::PathBuf;
+use std::path::{PathBuf, Path};
+use std::env;
+use std::process::Command;
 
 #[cfg(target_os = "linux")]
 const FEATURE_CHECKS_PATH: &str = "feature_checks";
+
+const C_DEPS_PATH: &str = "c_deps";
 
 /// Simple feature check, returning `true` if we have the feature.
 ///
@@ -56,6 +60,24 @@ fn feature_check(filename: &str) -> bool {
     check_build.file(path).try_compile("check_perf_pt").is_ok()
 }
 
+fn build_libipt() {
+    eprintln!("Building libipt...");
+    env::set_current_dir(&Path::new(C_DEPS_PATH)).unwrap();
+    let res = Command::new("make")
+        .arg("libipt")
+        .output()
+        .unwrap_or_else(|_| panic!("Fatal error when building libipt"));
+    if !res.status.success() {
+        eprintln!("libipt build failed\n>>> stdout");
+        eprintln!("stdout: {}", String::from_utf8_lossy(&res.stdout));
+        eprintln!("\n>>> stderr");
+        eprintln!("stderr: {}", String::from_utf8_lossy(&res.stderr));
+        panic!();
+    }
+    env::set_current_dir(&Path::new("..")).unwrap();
+}
+
+
 fn main() {
     let mut c_build = gcc::Build::new();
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -6,3 +6,14 @@ This library only supports Intel Processor Trace, but in the future we hope to
 support alternatives such as Arm's CoreSight.
 
 **This is experimental code.**
+
+## Building
+
+Building is performed by Cargo. In short, most users will want:
```
ltratt: I think we can assume that everyone knows that this needs to be built with cargo, so we can remove this sentence and the example.

build.rs
```diff
@@ -56,17 +60,53 @@ fn feature_check(filename: &str) -> bool {
     check_build.file(path).try_compile("check_perf_pt").is_ok()
 }
 
+fn build_libipt() {
+    eprintln!("Building libipt...");
+    env::set_current_dir(&Path::new(C_DEPS_PATH)).unwrap();
+    let res = Command::new("make")
+        .arg("libipt")
+        .output()
+        .unwrap_or_else(|_| panic!("Spawning the libipt build failed"));
```
ltratt: I'm not sure this error is accurate: other things than spawning could have failed? How about just "Fatal error when building libipt."?

build.rs
```diff
@@ -56,17 +60,53 @@ fn feature_check(filename: &str) -> bool {
...
     // Check for a new enough Perf to support Intel PT.
     #[cfg(target_os = "linux")] {
         if feature_check("check_perf_pt.c") {
             c_build.file("src/backends/perf_pt/perf_pt.c");
+            c_build.file("src/backends/perf_pt/decode.c");
+
+            // Decide whether to build our own libipt.
+            println!("cargo:rerun-if-env-changed=WITH_SYSTEM_IPT");
+            println!("cargo:rerun-changed={}", C_DEPS_PATH);
+            match env::var("WITH_SYSTEM_IPT") {
```
ltratt: Up to you, but this would be 1 line shorter, and 1 level of indentation shallower, if you used `if let` and `else`.

c_deps/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+all: libipt
+
+libipt: ${INST_DIR}/bin/ptdump
+
+${INST_DIR}:
+	install -d ${INST_DIR}/bin
+
+#
+# Fetch targets
```
ltratt: Can I suggest we remove the `#` lines above and below (i.e. have 1 blank line above and below this line)? I'd actually find it makes the important line pop out better.

src/backends/perf_pt/decode.c
```diff
@@ -0,0 +1,47 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include <intel-pt.h>
+
+void
+perf_pt_start_decode(void *buf) {
+    // XXX This is function body is only here to check for correct libipt
```
ltratt: Delete the first "is"?

README.md
```diff
@@ -6,3 +6,14 @@ This library only supports Intel Processor Trace, but in the future we hope to
...
+
+```
+cargo build --release
+```
+
+You can set `WITH_SYSTEM_IPT=...` to specify a path to a system libipt to use.
```
ltratt: Although it's a bit subjective `WITH_SYSTEM_` feels like a weird prefix. Why "with"? Why does the "system" have to have an ipt install? I think what you're really saying is "if you want to use your own version of libibt (whether or not it's installed system wide or not) set this var". How about just `IPT_PATH` or similar?

ltratt:
OK, several bike-shedding comments (sorry) but nothing substantial.

README.md
vext01: It's standard wording for build systems, e.g. in tex live we have:

```
			--with-system-freetype2 \
			--with-system-gd \
			--with-system-libpng \
			--with-system-ncurses \
			--with-system-pnglib \
			--with-system-zlib \
...
```

"system" means the one provided elsewhere, usually the package manager.

ltratt: It's a standard for configure switches, I agree, but not the environment variables. e.g. we just `CC` `CXX` and so on.

vext01: How about `IPT_PREFIX`. That actually better conveys that I expect there to be `include` and `lib` subdirs.

ltratt: It's better than what's before, but `PREFIX` makes me think of the part of a filename before a "."...

vext01:
Comments addressed.

ltratt:
Looks good to me. If you squash, I can merge.

vext01:
splat.

PR #42, opened 2018-03-15
Ensure traces start with a `PSB+` packet sequence.

### LINKED ISSUE #41: Really small traces will fail to decode.
I've noticed that in the simple example, if `iters` is less than about 55, then often the trace will not decode properly and libipt flags "expected tracing turned on event".

Not sure why this might be.

vext01:
This is the first of a few fixes coming in to make the trace decoding branch work.

This fixes #41. Please see the issue for a description of why this change is required.

The doc test change is weird. Only after the first commit does it crash. It should have been crashing all along. For what reason, I don't know. Anyway, it's fixed now.

OK?

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -570,8 +570,9 @@ perf_pt_free_tracer(struct tracer_ctx *tr_ctx) {
     if (tr_ctx->stop_fds[0] != -1) {
         close(tr_ctx->stop_fds[0]);
     }
-    if (tr_ctx->perf_fd) {
+    if (tr_ctx->perf_fd >= 0) {
         close(tr_ctx->perf_fd);
+        tr_ctx->perf_fd = -1;
     }
     if (tr_ctx != NULL) {
         free(tr_ctx);
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -39,6 +39,7 @@ use libc::{pid_t, c_void, size_t, geteuid, malloc, free};
 use errors::HWTracerError;
 use std::fs::File;
 use std::io::Read;
+use std::ptr;
 use Tracer;
 use util::linux_gettid;
 #[cfg(debug_assertions)]
@@ -166,14 +167,14 @@ impl PerfPTConf {
 
 /// A tracer that uses the Linux Perf interface to Intel Processor Trace.
 pub struct PerfPTTracer {
+    // The configuration for this tracer.
+    config: PerfPTConf,
     // Opaque C pointer representing the tracer context.
     tracer_ctx: *mut c_void,
     // The state of the tracer.
     state: TracerState,
     // The trace currently being collected, or `None`.
     trace: Option<Box<PerfPTTrace>>,
-    // The starting trace buffer size for new traces.
-    new_trace_bufsize: size_t,
 }
 
 impl PerfPTTracer {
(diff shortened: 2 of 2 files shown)


vext01:
Branch synced.

src/backends/perf_pt/collect.c
```diff
@@ -570,8 +570,9 @@ perf_pt_free_tracer(struct tracer_ctx *tr_ctx) {
     if (tr_ctx->stop_fds[0] != -1) {
         close(tr_ctx->stop_fds[0]);
     }
-    if (tr_ctx->perf_fd) {
+    if (tr_ctx->perf_fd >= 0) {
         close(tr_ctx->perf_fd);
+        tr_ctx->perf_fd = -1;
```
ltratt: Do we know for sure that valid fds have to be `>=` 0?

src/backends/perf_pt/mod.rs
```diff
@@ -277,12 +275,20 @@ impl Tracer for PerfPTTracer {
             return Err(HWTracerError::TracerAlreadyStarted);
         }
 
+        // At the time of writing, we have to use a fresh Perf file descriptor to ensure traces
```
ltratt: I think the second sentence ("this is required") is redundant?

ltratt:
Am an I an idiot if I say that I can't see where this PR closes a file handle?

src/backends/perf_pt/collect.c
vext01: Since open returns -1 on error...

ltratt: Ah, even more to the point the OpenBSD `open` page says "If successful, open() returns a non-negative integer, termed a file descriptor." Just checking!

vext01:
It's here:
https://github.com/softdevteam/hwtracer/blob/5bb7c9a3b9727c22f2100be3c911ec3f22d9e18e/src/backends/perf_pt/collect.c#L573-L575

And the call to `perf_pt_free_tracer` was moved here:
https://github.com/softdevteam/hwtracer/blob/5bb7c9a3b9727c22f2100be3c911ec3f22d9e18e/src/backends/perf_pt/mod.rs#L311-L313

i.e. when we stop tracing, not when we destroy the tracer.

ltratt:
Aha, OK.

src/backends/perf_pt/mod.rs
vext01: I think it's OK as-is. A trace not starting with a `PSB+` is a valid trace, but it may not decode.

ltratt: How can it be valid if it can't be decoded?

vext01: There are many levels. The packets decoder would succeed, but the instruction and block decoders would not.

vext01: So it's a valid trace at the packet level, if you see what I mean?

ltratt: Ah, so maybe we need to say which decoder wouldn't work in the comment?

vext01:
Try this.

src/backends/perf_pt/mod.rs
```diff
@@ -276,8 +276,8 @@ impl Tracer for PerfPTTracer {
         }
 
         // At the time of writing, we have to use a fresh Perf file descriptor to ensure traces
-        // start with a `PSB+` packet sequence. This is required for correct decoding. For this
-        // reason we have to re-initialise for each new tracing session.
+        // start with a `PSB+` packet sequence. This is required for correct instruction- and
```
ltratt: On second reading I know why the "-" is at the end of "instruction" -- but it's pretty distracting. I think it's safe enough to just drop it.

vext01: We even have enough space on the line to repeat `-level`. Looks good?

ltratt:
Yep, let's squash.

vext01:
splat.

ltratt:
Should the 1st and 3rd commits be squashed together?

vext01:
I think they are separate issues. The close logic was wrong in master too.

ltratt:
OK.

PR #43, opened 2018-03-15
perf_event_open could return EBUSY.

vext01:
This fixes intermittent test failures when tests are run in parallel.

Should we include a wait, or a maximum attempt limit?

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -57,9 +57,13 @@
 #include <stdbool.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <time.h>
 
 #define SYSFS_PT_TYPE   "/sys/bus/event_source/devices/intel_pt/type"
 #define MAX_PT_TYPE_STR 8
+
+#define MAX_OPEN_PERF_TRIES  500
+#define OPEN_PERF_WAIT_NSECS 1000 * 20
 
 #ifndef INFTIM
 #define INFTIM -1
@@ -264,8 +268,18 @@ open_perf(pid_t target_tid) {
     // No skid.
     attr.precise_ip = 3;
 
-    // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    // Acquire file descriptor through which to talk to Intel PT. This syscall
+    // could return EBUSY, meaning another process or thread has locked the
+    // Perf device.
+    struct timespec wait_time = {0, OPEN_PERF_WAIT_NSECS};
+    for (int tries = MAX_OPEN_PERF_TRIES; tries > 0; tries--) {
+        ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+        if ((ret == -1) && (errno == EBUSY)) {
+            nanosleep(&wait_time, NULL); // Doesn't matter if this is interrupted.
+        } else {
+            break;
+        }
+    }
 
 clean:
     if ((pt_type_file != NULL) && (fclose(pt_type_file) == -1)) {


src/backends/perf_pt/collect.c
```diff
@@ -265,7 +265,22 @@ open_perf(pid_t target_tid) {
     attr.precise_ip = 3;
 
     // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    //
```
ltratt: Delete the blank comment line?

src/backends/perf_pt/collect.c
```diff
@@ -265,7 +265,22 @@ open_perf(pid_t target_tid) {
...
+    // This syscall could return EBUSY, meaning another process or thread has
+    // locked the Perf device. We spin around until we get an fd.
+    while (true) {
```
ltratt: `loop {`

ltratt: Gah, wrong language. Mea culpa!

src/backends/perf_pt/collect.c
```diff
@@ -265,7 +265,22 @@ open_perf(pid_t target_tid) {
...
+        ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+        if (ret == -1) {
+            if (errno == EBUSY) {
+                continue;
+            } else {
+                ret = -1;
```
ltratt: `ret` is -1 by definition at this point so this assignment is unneeded?

src/backends/perf_pt/collect.c
```diff
@@ -265,7 +265,22 @@ open_perf(pid_t target_tid) {
     attr.precise_ip = 3;
 
     // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    //
+    // This syscall could return EBUSY, meaning another process or thread has
+    // locked the Perf device. We spin around until we get an fd.
+    while (true) {
+        ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+        if (ret == -1) {
```
ltratt: I think this entire `if` is equivalent to:

```c
if (ret == -1 && errno == EBUSY) {
    continue;
} else {
    break;
}
```

ltratt:
I agree we probably need a timeout to capture weird situations. In your experience so far, does this tend to succeed on the 2nd call, or does it require lots of retrying? What we probably want to do is work out what the practical maximum we see is, multiply that by some arbitrary constant (2,4,10, whatever) and set that as our limit.

src/backends/perf_pt/collect.c
vext01: It's intentional. It's spacing the following text from the above (untouched) comment line.

src/backends/perf_pt/collect.c
vext01: Hahahah!

src/backends/perf_pt/collect.c
ltratt: Just make it a blank line then without the `//`?

src/backends/perf_pt/collect.c
vext01: If we are into saving keystrokes, we could do `for (;;)`.

ltratt: I think `while (true)` is clearest myself.

src/backends/perf_pt/collect.c
vext01: Correct, good spot.

src/backends/perf_pt/collect.c
vext01: Where do you catch the other `-1` cases that `perf_event_open` could return?

ltratt: I don't, but nor does the commit: if it gets `-1` and it's not `EBUSY` it `goto clean`s.

vext01:
It happens quite rarely. I usually have to run the tests 4 or 5 times to even trigger this case. I'm hesitant to use a `sleep` because this is a performance critical code path, but we could give up after say 10/30/100 times? I'll try `10` and run tests lots and see what comes up.

ltratt:
It doesn't matter *too* much if we `sleep` after `EBUSY` because by definition we're already going to be delayed and it's not a case that will happen too often. I agree that finding out how many times round the loop we need to go to deal with an `EBUSY` is a good plan.

src/backends/perf_pt/collect.c
ltratt: In this PR or ...? Assuming you mean in a later PR, I'd suggest we simplify it as I suggested for the time being because then when you do come to change it, it'll be clearer what's going on. At the moment it is a bit confusing.

vext01:
Oh dear:

```
tries: 10
tries: 9
tries: 8
tries: 7
tries: 6
tries: 5
tries: 4
tries: 3
tries: 2
tries: 1
test backends::perf_pt::tests::test_repeated_tracing ... FAILED
```

Amazingly it can take more than 1000 tries, so I suppose that's a good excuse for a sleep!

How long?

ltratt:
1000 tries isn't likely to be very long in seconds. I would sleep some tiny number of microseconds. We'll still need some sort of timeout on that (e.g. let's say we sleep for 10 microseconds, we might allow 100 retries).

vext01:
* With a 10us sleep we easily max out 100 tries.
 * With a 20us sleep we often retry as many as 80 times.

So how about 20us/500 tries?

vext01:
I should mention that this scenario is less likely to arise in real tracing situations. The tests fire off a ton of tracing sessions in parallel, which is somewhat an artificial scenario. Nevertheless, we do need to handle it.

ltratt:
Works for me.

vext01:
Try this.

src/backends/perf_pt/collect.c
vext01: Er, well it's part of the same comment.

It's supposed to read like:
```
// Do this thing.
//
// Explain why.
// blah blah.
```

vext01: ( I don't feel strongly, so if you really want to kill the line...)

src/backends/perf_pt/collect.c
```diff
@@ -265,7 +265,19 @@ open_perf(pid_t target_tid) {
...
     // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    //
+    // This syscall could return EBUSY, meaning another process or thread has
+    // locked the Perf device.
+    for (int tries = 500; tries > 0; tries--) {
+        printf("tries: %d\n", tries);
+        ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+        if ((ret == -1) && (errno == EBUSY)) {
+            usleep(20); // Doesn't matter if usleep fails.
```
ltratt: Should we use `nanosleep`? My man page says `usleep is obsoleted` so I don't know if it's completely portable or not whereas `nanosleep` is in POSiX.

src/backends/perf_pt/collect.c
ltratt: I've never seen anyone except Knuth do this, and that whole literate programming thing never caught on ;) I'd prefer we killed either the line or at least the `//`. Sorry for nitpicking!

src/backends/perf_pt/collect.c
vext01: Yes, even the Linux man page agrees:
```
       4.3BSD,  POSIX.1-2001.   POSIX.1-2001  declares this function obsolete;
       use nanosleep(2) instead.  POSIX.1-2008 removes  the  specification  of
       usleep().
```

Fix coming.

src/backends/perf_pt/collect.c
vext01: What can I say. People of our calibre (me + Knuth)...

Fix coming.

vext01:
I think this is good to squash now?

src/backends/perf_pt/collect.c
```diff
@@ -264,8 +265,20 @@ open_perf(pid_t target_tid) {
...
     attr.precise_ip = 3;
 
-    // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    // Acquire file descriptor through which to talk to Intel PT. This syscall
+    // could return EBUSY, meaning another process or thread has locked the
+    // Perf device.
+    struct timespec wait_time = {0, 1000 * 20}; // 20 microseconds.
+    for (int tries = 500; tries > 0; tries--) {
+        printf("tries: %d\n", tries);
```
vext01: Argh!

ltratt: Debugging `printf`.

src/backends/perf_pt/collect.c
```diff
@@ -264,8 +265,20 @@ open_perf(pid_t target_tid) {
     // No skid.
     attr.precise_ip = 3;
 
-    // Acquire file descriptor through which to talk to Intel PT.
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    // Acquire file descriptor through which to talk to Intel PT. This syscall
+    // could return EBUSY, meaning another process or thread has locked the
+    // Perf device.
+    struct timespec wait_time = {0, 1000 * 20}; // 20 microseconds.
+    for (int tries = 500; tries > 0; tries--) {
```
ltratt: Can we pull this and the constant above into `#define`s at the top of the file?

vext01:
Killed a debug line.

src/backends/perf_pt/collect.c
vext01: I did consider this, along with the timeout too. One moment.

vext01:
Try this?

src/backends/perf_pt/collect.c
```diff
@@ -264,8 +268,19 @@ open_perf(pid_t target_tid) {
...
-    ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+    // Acquire file descriptor through which to talk to Intel PT. This syscall
+    // could return EBUSY, meaning another process or thread has locked the
+    // Perf device.
+    struct timespec wait_time = {0, OPEN_PERF_WAIT_NSECS};
+    for (int tries = MAX_OPEN_PERF_TRIES; tries > 0; tries--) {
+        ret = syscall(SYS_perf_event_open, &attr, target_tid, -1, -1, 0);
+        if ((ret == -1) && (errno == EBUSY)) {
+            nanosleep(&wait_time, NULL); // Doesn't matter if this is interrupted.
+            continue;
```
ltratt: This line is redundant.

vext01:
Hopefully now? :P

ltratt:
Yep, please squash.

[I know it took a while, but it's much nicer now!]

vext01:
It's fine, the code is better for it.

Splat

vext01:
Next up #42.

Issue #3, opened 2017-11-28
Proper error handling


vext01:
If we get to a point where this is definitely useful, we should add proper error handling. Many places in the code panic at the first hint of trouble.

ltratt:
One general comment: it's hard (as in: annoying) to add `Result` types retrospectively, so if you do think you'll need it, you might not want to wait too long...

vext01:
right, i'll bump the priority of this then.

vext01:
Fixed in #11

PR #44, opened 2018-03-15
Bits and pieces

vext01:
These are some bits that I've extracted from the decode branch as they are unrelated. If we do them as a separate PR I hope the decoder branch will be easier to review.

Looks OK?

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -58,6 +58,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <time.h>
+#include <stdatomic.h>
 
 #define SYSFS_PT_TYPE   "/sys/bus/event_source/devices/intel_pt/type"
 #define MAX_PT_TYPE_STR 8
@@ -128,7 +129,7 @@ struct tracer_thread_args {
 
 
 // Private prototypes.
-static bool read_aux(void *, __u64, __u64, __u64 *, struct perf_pt_trace *);
+static bool read_aux(void *, struct perf_event_mmap_page *, struct perf_pt_trace *);
 static bool poll_loop(int, int, struct perf_event_mmap_page *, void *,
                       struct perf_pt_trace *);
 static void *tracer_thread(void *);
@@ -143,16 +144,24 @@ bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx);
 /*
  * Read data out of the AUX buffer.
  *
- * Reads `size` bytes from `aux_buf` into `trace_buf`, updating `*trace_len`.
+ * Reads from `aux_buf` (whose meta-data is in `hdr`) into `trace`.
  *
  * Returns true on success and false otherwise.
  */
 static bool
-read_aux(void *aux_buf, __u64 size, __u64 head_monotonic, __u64 *tail_p,
-         struct perf_pt_trace *trace)
+read_aux(void *aux_buf, struct perf_event_mmap_page *hdr, struct perf_pt_trace *trace)
 {
+    // We use atomic loads and stores with orderings to avoid concurrency
+    // issues on the AUX state mutably shared between us and the kernel.
+    // Without these, for example the compiler might be able to load data
+    // from `aux_buf` before reading from the `hdr->aux_head`.
+    __u64 head_monotonic =
+            atomic_load_explicit((_Atomic __u64 *) &hdr->aux_head,
+                                 memory_order_acquire);
+    __u64 size = hdr->aux_size; // No atomic load. Not mutated in-kernel.
     __u64 head = head_monotonic % size; // Head must be manually wrapped.
-    __u64 tail = *tail_p;
+    __u64 tail = atomic_load_explicit((_Atomic __u64 *) &hdr->aux_tail,
+                                 memory_order_acquire);
 
     // First check we won't overflow the (destination) trace buffer.
     __u64 new_data_size;
@@ -175,9 +184,7 @@ read_aux(void *aux_buf, __u64 size, __u64 head_monotonic, __u64 *tail_p,
         memcpy(trace->buf, aux_buf, head);
     }
     trace->len += new_data_size;
-
-    // Update buffer tail, thus marking the space we just read as re-usable.
-    *tail_p = head;
+    atomic_store_explicit((_Atomic __u64 *) &hdr->aux_tail, head, memory_order_release);
     return true;
 }
 
@@ -205,12 +212,7 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
         }
 
         if ((pfds[0].revents & POLLIN) || (pfds[1].revents & POLLHUP)) {
-            // See <linux/perf_event.h> for why we need the asm block.
-            __u64 head = mmap_hdr->aux_head;
-            asm volatile ("" : : : "memory");
-
-            if (!read_aux(aux, mmap_hdr->aux_size, head, &mmap_hdr->aux_tail,
-                          trace)) {
+            if (!read_aux(aux, mmap_hdr, trace)) {
                 ret = false;
                 goto done;
             }
--- a/README.md
+++ b/README.md
@@ -12,3 +12,6 @@ support alternatives such as Arm's CoreSight.
 When running `cargo`, you can set `IPT_PATH=...` to specify a path to a system
 libipt to use. If this variable is absent, Cargo will download and build libipt
 for you.
+
+If you let Cargo download a libipt, you will have to set your `LD_LIBRARY_PATH`
+to `c_deps/inst/lib` before using hwtracer. This is due to a Cargo limitation.
--- a/build.rs
+++ b/build.rs
@@ -119,7 +119,6 @@ fn main() {
 
     // Additional circumstances under which to re-run this build.rs.
     println!("cargo:rerun-if-env-changed=IPT_PATH");
-    println!("cargo:rerun-if-env-changed=WITHOUT_IPT");
     println!("cargo:rerun-if-changed=src/util");
     println!("cargo:rerun-if-changed={}", C_DEPS_PATH);
     println!("cargo:rerun-if-changed=src/backends/perf_pt");
--- a/build.rs
+++ b/build.rs
@@ -60,6 +60,9 @@ fn feature_check(filename: &str) -> bool {
     check_build.file(path).try_compile("check_perf_pt").is_ok()
 }
 
+// XXX Currently no way to clean the c_deps dir at `cargo clean` time:
+// https://github.com/rust-lang/cargo/issues/572
+
 fn build_libipt() {
     eprintln!("Building libipt...");
     env::set_current_dir(&Path::new(C_DEPS_PATH)).unwrap();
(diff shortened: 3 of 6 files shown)


src/backends/perf_pt/collect.c
```diff
@@ -177,6 +177,8 @@ read_aux(void *aux_buf, __u64 size, __u64 head_monotonic, __u64 *tail_p,
     trace->len += new_data_size;
 
     // Update buffer tail, thus marking the space we just read as re-usable.
+    // See <linux/perf_event.h> for why we need the memory barrier.
```
ltratt: That doesn't explain why we need a barrier, nor why we're using `__sync_synchronize` rather than `barrier` as in the header. Also the example in the header uses two barriers. Am I missing out on something obvious?

vext01: It's not obvious, just a bit of a mess ;)

The part of the header i'm referring to is:
```
    /*                                                                          
     * Control data for the mmap() data buffer.                                 
     *                                                                          
     * User-space reading the @data_head value should issue an smp_rmb(),       
     * after reading this value.                                                
     *                                                                          
     * When the mapping is PROT_WRITE the @data_tail value should be            
     * written by userspace to reflect the last read data, after issueing       
     * an smp_mb() to separate the data read from the ->data_tail store.        
     * In this case the kernel will not over-write unread data.                 
     *                                                                          
     * See perf_output_put_handle() for the data ordering.                      
     *                                                                          
     * data_{offset,size} indicate the location and size of the perf record     
     * buffer within the mmapped area.                                          
     */
```

So we are talking about `smp_rmb()` and `smp_rb()` (I'd not heard of `barrier()` until you mentioned it, but that part of the header is talking about something different from reading the AUX buffer. FWIW there is no man page for `barrier()` so it's probably in the kernel somewhere). The `smp_*` functions are not exposed to userspace (this is a kernel header), so we have to do something different. The `asm` block is something that Mario pointed me to, whereas the `__sync_synchronise()` call is something I found here:
https://stuff.mit.edu/afs/sipb/contrib/linux/samples/bpf/trace_output_user.c

However! @davmac314 has since pointed out that C11 has native memory barrier support in the form of `atomic_thread_fence()`. I think we have been using C99 up until this point. I could try add a `-std=C11` (or whatever) flag and seeing if anything explodes.

Do you have any opinion on any of this?

vext01: I should also point out, the pasted section of the header above is talk about the `data` buffer, but later on it says:
```
     * Ring buffer pointers aux_{head,tail} have the same semantics as    
     * data_{head,tail} and same ordering rules apply.    
```

ltratt: If using C11 helps us write better code, I'm all for it.

There's a lot of low-level detail in your messages. Let me see if I can condense them into a high-level story. I think it's the following (tell me where I'm wrong!). The docs tell us to use `smp_mb` after reading the value, but that's not available to use; `__sync_synchronise` is a synonym for `smp_mb`; and `atomic_thread_fence` is a synonym for `__sync_synchronise`.

vext01: I'm not totally sure myself. It seems to me that `atomic_thread_fence` is an acquire/release-style memory barrier (but does not couple with an atomic read/write).

I think we have three options:

1) Remove the syncronisation bits and put a comment saying "this is x86_64 only code, so we don't need to use memory barriers".

2) Copy what the Perf does:

```
  static inline u64 auxtrace_mmap__read_head(struct auxtrace_mmap *mm)    
  {    
      struct perf_event_mmap_page *pc = mm->userpg;    
  #if BITS_PER_LONG == 64 || !defined(HAVE_SYNC_COMPARE_AND_SWAP_SUPPORT)    
      u64 head = READ_ONCE(pc->aux_head);    
  #else    
      u64 head = __sync_val_compare_and_swap(&pc->aux_head, 0, 0);    
  #endif    
      
      /* Ensure all reads are done after we read the head */    
      rmb();    
      return head;    
  }
```

```
  static inline void auxtrace_mmap__write_tail(struct auxtrace_mmap *mm, u64 tail)
  {    
      struct perf_event_mmap_page *pc = mm->userpg;    
  #if BITS_PER_LONG != 64 && defined(HAVE_SYNC_COMPARE_AND_SWAP_SUPPORT)    
      u64 old_tail;    
  #endif    

      /* Ensure all reads are done before we write the tail out */
      mb();
  #if BITS_PER_LONG == 64 || !defined(HAVE_SYNC_COMPARE_AND_SWAP_SUPPORT)
      pc->aux_tail = tail;
  #else
      do {
          old_tail = __sync_val_compare_and_swap(&pc->aux_tail, 0, 0);
      } while (!__sync_bool_compare_and_swap(&pc->aux_tail, old_tail, tail));
  #endif
  }
```

3) Convert 2) to C11 atomic primitives.

I think 2 is less error prone than 3, but you may also find 1 viable.

vext01: (Note the use of atomics in 2, which is currently missing from my implementation, assuming we don't go for 1).

vext01: And we'd have to convert the `smp_*` bits if we copy 2. Let me see if I can find a userspace example.

ltratt: If we're not sure if #3 is correct, then we can't possibly do it.

One thing we need to think about is that fences aren't just about run-time semantics: they also effect compile-time reorderings. So we might need a fence (or equivalent) here simply to stop the compiler fiddling with our code, even on x86_64.

davmac314:
@ltratt:
> __sync_synchronise is a synonym for smp_mb; and atomic_thread_fence is a synonym for __sync_synchronise

That's how I understand it. AFAICS you need a barrier to prevent any re-ordering of data going to the head/tail pointers and data going into the buffer itself. I.e. you don't want:

1. read head pointer
2. read buffer[head]  // oops, we saw the update to head before we saw the update to buffer

In C11 terms reading the head pointer is an acquire operation, so the simplest way is to read the head pointer via an atomic load tagged with "acquire" semantics.

ltratt:
Thanks Davin!

ltratt:
Given that an "atomic int" is just a normal int that we happen to use atomic operations on, I don't see why a cast wouldn't work. All that said, a fence might be easier, as at least it's only in a single place.

vext01:
Note that a fence is only one half of the solution, to do it by the book (tm) you would still need atomic ops.

> Given that an "atomic int" is just a normal int that we happen to use atomic operations on

Not sure.

davmac314:
> as that would require the aux_head and aux_tail to have been defined _Atomic, whereas they are just plain __u64s

That doesn't actually stop you using atomic operations (while I'm not entirely sure what the language standard says about it). Eg https://godbolt.org/g/X9hSfy

ltratt:
@vext01 None of the docs you posted mention atomic ints... So I'm not sure where that requirement comes from.

ltratt:
Re: atomic ops. The only problem with using atomic operations on things that aren't declared as atomic is that if function F accesses int I atomically but function G access int I non-atomically, then G's going to sometimes see odd effects. But really all that's going on there is a standard synchronisation cock-up.

vext01:
@ltratt If you look at the perf code I pasted, all of the memory barriers are coupled with an atomic load/store.

ltratt:
Well, I see CAS operations, but you're right: if someone's CASing in one thread, we need to read atomically in our thread.

vext01:
So perhaps I will try to emulate, as closely as possible, what the perf devs have done in their own code. I don't think we can go far wrong with that strategy.

davmac314:
FWIW I think that compare-and-swap is probably used because they figure it might be the only atomic operation supported for the data type. Note that it's only used if BITS_PER_LONG != 64. And as compare-and-swaps go it's not really correct (or at least, it's excessive; the loop is pointless; if the value changes in another thread why not just let it change rather than keep looping until we can successfully clobber its value? Someone is always going to be clobbering someone else).

TLDR - the c-a-s is probably only used as an atomic load on platforms that don't have a 64-bit atomic load (but do presumably have a 64-bit c-a-s, which is more common).

vext01:
Since processor trace is pretty new, I think we can assume a modern Intel chip with all of the atomic ops.

davmac314:
> Since processor trace is pretty new, I think we can assume a modern Intel chip with all of the atomic ops.

Then I would personally use C11 and be done with it. Your "read_head" operation then goes from several lines of code, some of which look dubious anyway, to a single line which is guaranteed (bar compiler / processor bugs) to do the right thing. Similarly for "write_tail".

vext01:
I think this is correct. I didn't have to add any compiler flags.

ltratt:
I might have misunderstood Davin, but is he saying that a fence alone is enough and we don't need the atomic operations as well?

davmac314:
A compiler fence would be enough on x86-64, but, using an atomic operation is arguably more correct (and doesn't incur any penalty on x86-64; it's compiled to a standard non-atomic load or store).

ltratt:
More correct? I'm old-fashioned: either it's correct or it isn't ;)

vext01:
If we were bothered about portability? Right?

Without the atomic ops, writes could sit in a local cache, and reads may not hit your local cache.

davmac314:
> either it's correct or it isn't ;)

It's only correct, by the language standard, _with_ the atomic operation - which is why I'd say use the atomic operation. But realistically very few C programs are correct by the language standard...

vext01:
FWIW, a disassembly of `read_aux` doesn't show anything unusual. I *think* the atomic ops boil down to a no-op at the hardware level.

vext01:
Or rather, they disappear  (on x86_64).

vext01:
So, can I squash? Or are there more questions about the atomics?

vext01:
Sorry, one additional `build.rs` tweak.

src/backends/perf_pt/collect.c
```diff
@@ -144,16 +144,21 @@ bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx);
...
+ * Reads from `aux_buf` (whose meta-data is in `hdr`) into `trace`.
  *
  * Returns true on success and false otherwise.
  */
 static bool
-read_aux(void *aux_buf, __u64 size, __u64 head_monotonic, __u64 *tail_p,
-         struct perf_pt_trace *trace)
+read_aux(void *aux_buf, struct perf_event_mmap_page *hdr, struct perf_pt_trace *trace)
 {
+    // atomic loads not strictly needed on x86_64, but we play it by the book.
```
ltratt: It's worth noting that the atomic reads are doing two things: they have a run-time effect (but, I agree, on x86_64, atomic/non-atomic reads are the same at run-time); but they have a compile-time effect (they prevent the compiler moving things around). So I don't think this line is true: without using atomics, it's possible the compiler could change the semantics of our program.

vext01: It's true that these ordered atomic ops impose compiler barriers. Let me revise the comment.

ltratt: I think it can just be deleted...

vext01:
Try that @ltratt

ltratt:
The first part of the comment is still, IMHO, misleading. They *are* strictly needed, just at compile-time, not run-time. IMHO, if you want a comment here, it would be most useful to say *why* we need to prevent reorderings (i.e. what invariants are we trying to maintain).

vext01:
No, atomic operations and memory barriers are completely separate things. C11 bundles them together because they are often used together.

I could extend the comment to talk about why we need the barriers though.

ltratt:
Atomic operations implicitly include barriers (well, assuming you specify a barrier!), so you can't really talk about atomics without barriers IMHO. That's why I find the first part of the comment misleading, even though I understand where you're coming from. It would be clearer to say something like "We need the acquire barrier that comes with an atomic read because XYZ".

vext01:
Yes, but we could have just used a compiler barrier (and no atomics) and gotten the same behaviour on x86_64.

Anyway I agree that we could take a more objective approach on this comment.

vext01:
Uh, I *think* that is right, but I'm not super-confident, so please check.

I have to admit I find this stuff *really* hard to think about.

vext01:
> Atomic operations implicitly include barriers.

Only at the language level.

vext01:
So I hope I can squash now?

ltratt:
Time to squash!

vext01:
Splat.

PR #45, opened 2018-03-16
Remove analyse.mk.

vext01:
I realised earlier this week that when run under valgrind, the CPU
reports not supporting PT, so we end up leak checking only the dummy
backend. It hardly seems worth it.

[Valgrind is an emulator after all].


ltratt:
This looks fine, but can you rebase against master?

vext01:
Branch synced.

PR #46, opened 2018-03-21
Block decoding support.

### LINKED ISSUE #28: Decode packet stream to get control flow.
Depends upon #27

vext01:
This adds support for block decoding into the perf_pt backend.

It's a large change and I'm pretty certain you will be able to find some nits! Let the discussion begin!

Anything `XXX` is either Rust/Cargo bugs/limitations, or something which I think can be tackled later in later PRs.

Fixes #28

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -80,6 +80,24 @@ fn build_libipt() {
     env::set_current_dir(&Path::new("..")).unwrap();
 }
 
+// We always fetch libipt regardless of if we will build our own libipt. This is becuase there are
+// a couple of private CPU configuration files that we need to borrow from libipt.
+fn fetch_libipt() {
+    eprintln!("Fetch libipt...");
+    env::set_current_dir(&Path::new(C_DEPS_PATH)).unwrap();
+    let res = Command::new("make")
+        .arg("processor-trace") // target just fetches the code.
+        .output()
+        .unwrap_or_else(|_| panic!("Fatal error when fetching libipt"));
+    if !res.status.success() {
+        eprintln!("libipt fetch failed\n>>> stdout");
+        eprintln!("stdout: {}", String::from_utf8_lossy(&res.stdout));
+        eprintln!("\n>>> stderr");
+        eprintln!("stderr: {}", String::from_utf8_lossy(&res.stderr));
+        panic!();
+    }
+    env::set_current_dir(&Path::new("..")).unwrap();
+}
 
 fn main() {
     let mut c_build = gcc::Build::new();
@@ -90,6 +108,10 @@ fn main() {
             c_build.file("src/backends/perf_pt/collect.c");
             c_build.file("src/backends/perf_pt/decode.c");
 
+            // XXX At the time of writing you can't conditionally build C code for tests in a build
+            // script: https://github.com/rust-lang/cargo/issues/1581
+            c_build.file("src/backends/perf_pt/test_helpers.c");
+
             // Decide whether to build our own libipt.
             if let Ok(val) = env::var("IPT_PATH") {
                 let mut inc_path = PathBuf::from(val.clone());
@@ -97,13 +119,22 @@ fn main() {
                 c_build.include(inc_path);
                 c_build.flag(&format!("-L{}/lib", val));
                 println!("cargo:rustc-link-search={}/lib", val);
+                println!("cargo:rustc-env=PTXED={}/bin/ptxed", val);
             } else {
                 build_libipt();
                 c_build.include(&format!("{}/inst/include/", C_DEPS_PATH));
                 c_build.flag(&format!("-L{}/inst/lib", C_DEPS_PATH));
                 println!("cargo:rustc-link-search={}/inst/lib", C_DEPS_PATH);
+                println!("cargo:rustc-env=PTXED={}/inst/bin/ptxed", C_DEPS_PATH);
             }
-            c_build.flag("-lipt");
+
+            // We borrow the CPU detection functions from libipt (they are not exposed publicly).
+            // If we built our own libipt above, then the fetch is a no-op.
+            fetch_libipt();
+            c_build.include(&format!("{}/processor-trace/libipt/internal/include", C_DEPS_PATH));
+            c_build.file(&format!("{}/processor-trace/libipt/src/pt_cpu.c", C_DEPS_PATH));
+            c_build.file(&format!("{}/processor-trace/libipt/src/posix/pt_cpuid.c", C_DEPS_PATH));
+
             println!("cargo:rustc-cfg=perf_pt");
             println!("cargo:rustc-link-lib=ipt");
 
@@ -122,4 +153,6 @@ fn main() {
     println!("cargo:rerun-if-changed=src/util");
     println!("cargo:rerun-if-changed={}", C_DEPS_PATH);
     println!("cargo:rerun-if-changed=src/backends/perf_pt");
+    println!("cargo:rerun-if-changed={}/processor-trace/libipt/src/pt_cpu.c", C_DEPS_PATH);
+    println!("cargo:rerun-if-changed={}/processor-trace/libipt/src/posix/pt_cpuid.c", C_DEPS_PATH);
 }
--- a/examples/simple_example.rs
+++ b/examples/simple_example.rs
@@ -38,9 +38,8 @@
 extern crate hwtracer;
 extern crate libc;
 
-use hwtracer::Tracer;
-use libc::getpid;
-
+use std::time::SystemTime;
+use hwtracer::{Tracer, Trace};
 use hwtracer::backends::DummyTracer;
 #[cfg(perf_pt)]
 use hwtracer::backends::PerfPTTracer;
@@ -65,36 +64,48 @@ fn tracer() -> Box<Tracer> {
     }
 }
 
+/// Prints the addresses of the first `qty` blocks in a trace along with it's name and
+/// computation result.
+fn print_trace(trace: &Box<Trace>, name: &str, result: u32, qty: usize) {
+    let count = trace.iter_blocks().count();
+    println!("{}: num_blocks={}, result={}", name, count, result);
+
+    for (i, blk) in trace.iter_blocks().take(qty) .enumerate() {
+       println!("  block {}: 0x{:x}", i, blk.unwrap().start_vaddr());
+    }
+    if count > qty {
+        println!("  ... {} more", count - qty);
+    }
+    println!("");
+}
+
+/// It's up to you to ensure the compiler doesn't move or optimise out the computation you intend
+/// to trace. Here we've used an uninlinable function. See also `test::black_box`.
+#[inline(never)]
+fn work() -> u32 {
+    let mut res = 0;
+    for _ in 0..50 {
+        // Computation which stops the compiler from eliminating the loop.
+        res += SystemTime::now().elapsed().unwrap().subsec_nanos();
+    }
+    res
+}
+
 /// Trace a simple computation loop.
 ///
 /// The results are printed to discourage the compiler from optimising the computation out.
 fn main() {
-    let mut res = 0;
-    let pid = unsafe { getpid() };
-
     let mut tracer = tracer();
-    tracer.start_tracing().unwrap_or_else(|e| {
-        panic!(format!("Failed to start tracer: {}", e));
-    });
 
-    let iters = 10_000;
-    for i in 1..iters {
-        res += i + pid;
+    for i in 1..4 {
+        tracer.start_tracing().unwrap_or_else(|e| {
+            panic!(format!("Failed to start tracer: {}", e));
+        });
+        let res = work();
+        let trace = tracer.stop_tracing().unwrap();
+        let name = format!("trace{}", i);
+        print_trace(&trace, &name, res, 10);
     }
-    let trace1 = tracer.stop_tracing().unwrap();
-    println!("Trace #1 = {:?}", trace1);
-    println!("result1: {}", res);
-
-    // Tracer instances can be re-used.
-    tracer.start_tracing().unwrap_or_else(|e| {
-        panic!(format!("Failed to start tracer: {}", e));
-    });
-    for i in 1..iters {
-        res -= i + pid;
-    }
-    let trace2 = tracer.stop_tracing().unwrap();
-    println!("Trace #2 = {:?}", trace2);
-    println!("result2: {}", res);
 
     // Tracers require manual clean up (since clean up could fail).
     tracer.destroy().unwrap();
--- a/src/backends/perf_pt/decode.c
+++ b/src/backends/perf_pt/decode.c
@@ -35,13 +35,310 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-#include <intel-pt.h>
+#define _GNU_SOURCE
 
-void
-perf_pt_start_decode(void *buf) {
-    // XXX This function body is only here to check for correct libipt linkage
-    // at this time.
-    (void) buf;
+#include <stdio.h>
+#include <intel-pt.h>
+#include <pt_cpu.h>
+#include <stdbool.h>
+#include <inttypes.h>
+#include <stdint.h>
+#include <link.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <assert.h>
+
+#include "perf_pt_private.h"
+
+struct load_self_image_args {
+    struct pt_image *image;
+    int vdso_fd;
+    char *vdso_filename;
+};
+
+// Private prototypes.
+static int drain_events(struct pt_block_decoder *, int);
+static bool load_self_image(struct load_self_image_args *);
+static int load_self_image_cb(struct dl_phdr_info *, size_t, void *);
+
+// Public prototypes.
+void *perf_pt_init_block_decoder(void *, uint64_t, int, char *, int *);
+bool perf_pt_next_block(struct pt_block_decoder *, int *, uint64_t *);
+void perf_pt_free_block_decoder(struct pt_block_decoder *);
+
+/*
+ * Get ready to retrieve the basic blocks from a PT trace using the code of the
+ * current process for control flow recovery.
+ *
+ * Accepts a raw buffer `buf` of length `len`.
+ *
+ * `vdso_fd` is an open file descriptor for the filename `vdso_filename`. This
+ * is where the VDSO code will be written. libipt will read this file lazily,
+ * so it's up to the caller to make sure this file lives long enough for their
+ * purposes.
+ *
+ * `*decoder_status` will be updated to reflect the status of the decoder after
+ * it has been synchronised.
+ *
+ * Returns a pointer to a configured libipt block decoder or NULL on error.
+ */
+void *
+perf_pt_init_block_decoder(void *buf, uint64_t len, int vdso_fd, char *vdso_filename,
+                           int *decoder_status) {
+    bool failing = false;
+
+    // Make a block decoder configuration.
     struct pt_config config;
-    pt_blk_alloc_decoder(&config);
+    memset(&config, 0, sizeof(config));
+    config.size = sizeof(config);
+    config.begin = buf;
+    config.end = buf + len;
+    config.flags.variant.block.end_on_call = 1;
+    config.flags.variant.block.end_on_jump = 1;
+
+    // Decode for the current CPU.
+    struct pt_block_decoder *decoder = NULL;
+    if (pt_cpu_read(&config.cpu) != 0) {
+        failing = true;
+        goto clean;
+    }
+
+    // Work around CPU bugs.
+    if ((config.cpu.vendor) && (pt_cpu_errata(&config.errata, &config.cpu) < 0)) {
+        failing = true;
+        goto clean;
+    }
+
+    // Instantiate a decoder.
+    decoder = pt_blk_alloc_decoder(&config);
+    if (decoder == NULL) {
+        failing = true;
+        goto clean;
+    }
+
+    // Sync the decoder.
+    *decoder_status = pt_blk_sync_forward(decoder);
+    if (*decoder_status == -pte_eos) {
+        // There were no blocks in the stream. The user will find out on next
+        // call to perf_pt_next_block().
+        goto clean;
+    } else if (*decoder_status < 0) {
+        failing = true;
+        goto clean;
+    }
+
+    // Build and load a memory image from which to recover control flow.
+    struct pt_image *image = pt_image_alloc(NULL);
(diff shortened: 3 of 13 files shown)


src/lib.rs
```diff
@@ -56,8 +103,12 @@ pub trait Trace: Debug {
     /// Dump the trace to the specified filename.
     ///
     /// The exact format varies per-backend.
-    #[cfg(debug_assertions)]
-    fn to_file(&self, filename: &str);
+    #[cfg(test)]
+    fn to_file(&self, file: &mut File);
+
+    /// Iterate over the blocks of the trace.
+    /// XXX if this were to return a `Result` then the `Block` type could disappear?
```
vext01: This is what we discussed on the walk in to King's today. I think this can be the next PR.

src/backends/perf_pt/mod.rs
```diff
@@ -334,11 +419,132 @@ impl Drop for PerfPTTracer {
...
+            trace = test_helpers::trace_closure(&mut tracer, &f);
+            expects = get_expected_blocks(&trace);
+            if expects.is_ok() {
+                break;
+            }
+        }
+        Some((tracer, trace, expects.unwrap()))
+    }
 
+    // XXX This isn't very flexible, as it only allows us to call test helpers with that one
```
vext01: And this will probably be the 2nd in line for future PRs.

vext01:
Unrelated network error on travis. kicked.

build.rs
```diff
@@ -80,6 +80,22 @@ fn build_libipt() {
     env::set_current_dir(&Path::new("..")).unwrap();
 }
 
+fn fetch_libipt() {
+    eprintln!("Fetch libipt...");
```
ltratt: Shouldn't this be "Build libipt" (and do we need the `...`?).

build.rs
```diff
@@ -115,11 +147,14 @@ fn main() {
         }
     }
     c_build.file("src/util/util.c");
+
```
ltratt: Extraneous blank line?

examples/simple_example.rs
```diff
@@ -65,36 +64,53 @@ fn tracer() -> Box<Tracer> {
     }
 }
 
+/// Prints the addresses of the first `qty` blocks in a trace along with it's name and
+/// computation result.
+fn print_trace(trace: &Box<Trace>, name: &str, result: u32, qty: usize) {
```
ltratt: This feels like a debug function to me? And should it be printing directly or building a string and returning it to the user? It feels a bit odd to pass a `name` to a function that prints it straight out to stdout.

examples/simple_example.rs
```diff
@@ -65,36 +64,53 @@ fn tracer() -> Box<Tracer> {
...
+    let count = trace.iter_blocks().count();
+    println!("{}: num_blocks={}, result={}", name, count, result);
+    let mut itr = trace.iter_blocks();
+    let mut end = false;
+    for i in 0..qty {
```
ltratt: I think you can simplify most of the code in this function down to something like:

```rust
trace.iter_blocks().take(qty).map(|x| println!("  block ...", ...));
if qty > count {
  println!("  ... {} more", ...);
}
```

examples/simple_example.rs
ltratt: Doh, ignore the "debug" comment -- I missed that this as an example :/

examples/simple_example.rs
```diff
@@ -65,36 +64,53 @@ fn tracer() -> Box<Tracer> {
...
+        }
+    }
+    if !end {
+        println!("  ... {} more", count - qty);
+    }
+    println!("");
+}
+
+/// It's up to you to ensure the compiler doesn't move or optimise out the computation you intend
+/// to trace. Here we've used an uninlinable function. See also `test::black`.
```
ltratt: `test::black_box`?

examples/simple_example.rs
```diff
@@ -65,36 +64,53 @@ fn tracer() -> Box<Tracer> {
...
+#[inline(never)]
+fn work() -> u32 {
+    let mut res = 0;
+    for _ in 0..50 {
+        // Computation which blocks the compiler from eliminating the loop.
```
ltratt: s/blocks/stops/

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+void perf_pt_free_decoder(struct pt_block_decoder *);
+
+/*
+ * Start decoding a PT trace.
+ *
+ * Accepts a raw buffer `buf` of length `len`.
+ *
+ * `vdso_fd` is an open file descriptor for the filename `vdso_filename`. This
+ * is where the VDSO code will be written. libipt will read this file lazily,
+ * so it's up to the caller to make sure this file lives long enough for their
```
ltratt: Does "the caller" mean "us"?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+ * purposes.
+ *
+ * `*decoder_status` will be updated to reflect the status of the decoder after
+ * it has been synchronised.
+ *
+ * Returns a pointer to a configured libipt block decoder or NULL on error.
+ *
+ * XXX This function assumes that we are decoding a trace for the current CPU
```
ltratt: This doesn't feel like an XXX to me: it's a fundamental assumption of the whole library, right?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+
+    // Decode for the current CPU.
+    struct pt_block_decoder *decoder = NULL;
+    if (pt_cpu_read(&config.cpu) != 0) {
+        failing = true;
+        goto clean;
+    }
+
+    // Work around CPU bugs.
+    if (config.cpu.vendor) {
```
ltratt: Use `&&` for the nested `if`?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+ */
+bool
+perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
+                   uint64_t *addr) {
+    // If there are events pending, drain them first.
+    *decoder_status = drain_events(decoder, *decoder_status);
+    if (*decoder_status < 0) {
+        // Error.
+        return false;
+    } else if (*decoder_status & pts_eos) {
```
ltratt: What's `pts_eos`? I'm not sure I understand the second and third branches of this `if` statement TBH.

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+
+    if (block.truncated == 1) {
+        // XXX A truncated block. This occurs when a block straddles a section
+        // boundary. In this case we may need some extra logic, but this should
+        // be a rare case.
+        return false;
+    }
+
+    if (block.ninsn == 0) {
+        // A block with zero instructions. Shouldn't happen.
```
ltratt: If it shouldn't happen, this should be an `assert`?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+/*
+ * Loads the libipt image `image` with the code of the current process.
+ *
+ * Returns true on success or false otherwise.
+ */
+static bool
+load_self_image(struct load_self_image_args *args)
+{
+    bool ret = dl_iterate_phdr(load_self_image_cb, args) == 0;
+    if (fsync(args->vdso_fd) == -1) {
```
ltratt: Can't we simplify this to `return fsync(...) != -1;`?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+ * on success and to continue iterating. See dl_iterate_phdr(3) for information
+ * on this interface.
+ */
+static int
+load_self_image_cb(struct dl_phdr_info *info, size_t size, void *data)
+{
+    ElfW(Phdr) phdr;
+    ElfW(Half) i;
+
+    (void) size;
```
ltratt: I'm not sure what this line is doing.

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+/*
+ * Dump the VDSO code into the open file descriptor `fd`, starting at `vaddr`
+ * and of size `len` into a temp file.
+ *
+ * Returns true on success or false otherwise.
+ */
+bool
+dump_vdso(int fd, uint64_t vaddr, size_t len)
+{
+    size_t wrote_so_far = 0;
```
ltratt: If you want a long variable name, how about `written`?

src/backends/perf_pt/decode.c
```diff
@@ -35,13 +35,317 @@
...
+    while (wrote_so_far != len) {
+        int wrote = write(fd, (void *) vaddr + wrote_so_far, len - wrote_so_far);
+        if (wrote == -1) {
+            return false;
+        }
+        wrote_so_far += wrote;
+    }
+
```
ltratt: I think all of the other functions up to this point haven't had a blank line before `return`.

src/backends/perf_pt/test_helpers.c
```diff
@@ -0,0 +1,144 @@
...
+
+/*
+ * Walks the program headers pushing ptxed `--raw` arguments for the current
+ * process into the supplied (pointer to a) Rust `AppendSelfPtxedArgs`
+ * instance.
+ */
+bool
+perf_pt_append_self_ptxed_raw_args(struct append_args *rs_args)
+{
+    if (dl_iterate_phdr(append_self_ptxed_raw_args_cb, rs_args) == 1) {
```
ltratt: Can just be `return dl_iterate_phdr(...) != 1;`

src/backends/perf_pt/test_helpers.c
```diff
@@ -0,0 +1,144 @@
...
+    return true;
+}
+
+/*
+ * Callback for perf_pt_append_self_ptxed_raw_args.
+ *
+ * `data` is a pointer to a Rust `AppendSelfPtxedArgs` instance.
+ */
+static int
+append_self_ptxed_raw_args_cb(struct dl_phdr_info *info, size_t size, void *data)
```
ltratt: Most of this function is the same as `load_self_image_cb` -- can/should we factor out the common bits?

src/lib.rs
```diff
@@ -94,31 +145,40 @@ enum TracerState {
...
-    fn work_loop() {
-        let mut res = unsafe { getpid() };
-        for _ in 0..100 {
-            res += 33;
+    // A loop that does some work that we can use to build a trace.
+    #[inline(never)]
+    pub fn work_loop(iters: u64) -> u64 {
+        let mut res = 0;
+        for _ in 0..iters {
+            // Computation which blocks the compiler from eliminating the loop.
```
ltratt: s/blocks/stops/

build.rs
vext01: There are separate build and fetch steps as we need to build a couple of files inside the libipt repo regardless of whether the user wants to use their own libipt or ours.

There's a comment later that tries to explain, I can move it up if you think it would be clearer.

vext01: `...` could go, yeah.

ltratt: Ah, yes, I see now. Moving the comment might help.

build.rs
vext01: It was intentional, as it splits configuring the C build from actually doing it. Can kill it if you like.

examples/simple_example.rs
vext01: Yeah, it's an example. Just doing something with the trace to offer the user some evidence that we actually did trace something :)

examples/simple_example.rs
vext01: Sounds good.

examples/simple_example.rs
vext01: oops, yes.

src/backends/perf_pt/decode.c
vext01: The caller of this function, which is "us" in a sense. This detail is not exposed to the consumers of our library.

ltratt: Right, so I think being clearer that it's "us" will help.

src/backends/perf_pt/decode.c
vext01: The CPU part is should probably be baked in.

With regards to "the current process", I don't see why our library couldn't be used more generally to trace other processes. We even offer a configuration option to choose an arbitrary `tid` to trace. But if we wanted to decode a trace for another process, we would have to offer an API to load the code for a process other than ourself.

ltratt: Maybe someone might do that as a PR, but I can't see that we'd ever want this?

src/backends/perf_pt/decode.c
vext01: `pts_eos` means "PT status: end of stream". In other words, there are no more blocks to iterate over. We return 0 which prompts the Rust iterator to return `None`.

The third case handles statuses that I've not yet considered. There's a list of possibilities here:
https://github.com/01org/processor-trace/blob/24982c1a6fce48f1e416461d42899805f74fbb26/libipt/include/intel-pt.h.in#L1064-L1074

* `pts_event_pending` should not occur, since this very call, by definition, should be draining the events from the stream.
* I don't think `pts_ip_suppressed` can happen as this call doesn't fetch any blocks.

None the less, I'd like to know if we get here. Would it be better as an assertion?

src/backends/perf_pt/decode.c
vext01: It could be yeah. Do you think that's better than returning an error?

src/backends/perf_pt/decode.c
vext01: We can!

vext01: Wait, what if `dl_iterate_phdr` fails, but fsync succeeds? We would return `true` ignoring the error.

src/backends/perf_pt/decode.c
vext01: It's a standard way to silence the compiler in the event of an unused argument. The argument is forced upon us by the signature of the `dl_iterate_phdr` callback.

src/backends/perf_pt/decode.c
vext01: I like it!

src/backends/perf_pt/test_helpers.c
vext01: Yeah. Don't know why I made that so complicated. It probably had some other code around it at some point.

src/backends/perf_pt/decode.c
ltratt: Yes, maybe an `assert` would be better because at least we'd know if they *can* happen.

src/backends/perf_pt/decode.c
ltratt: If it's not supposed to happen, then I think so.

src/backends/perf_pt/test_helpers.c
vext01: I did consider it, but thought it might be simpler to comprehend by having them separate like this. How strongly do you feel about it?

It might be a bit weird/confusing to have `void *data` be either a pointer to a Rust struct or a C `pt_image`.

src/backends/perf_pt/decode.c
ltratt: That's a good point. But should we do `fsync` if `dl_iterate_phdr` failed? Maybe we should really be doing something like:

```
return !((dl_iterate_phdr(load_self_image_cb, args) == 0) && (fsync(args->vdso_fd) != -1));
```

src/backends/perf_pt/decode.c
ltratt: Aha! Maybe doc that with `// Unused variable: silence compiler warnings` or similar?

src/backends/perf_pt/decode.c
vext01: Here's the question: Is hwtracer a library for our meta-tracer, or is it general purpose?

src/backends/perf_pt/test_helpers.c
ltratt: It's always a judgement call as to when to copy/paste and when to factor out. Your decision (I just wanted to flag it).

src/backends/perf_pt/decode.c
vext01: We can do, but after splitting the long line, it's probably not going to be pretty. I'll try it nonetheless.

src/backends/perf_pt/decode.c
ltratt: I am not generally a fan of putting in functionality for use cases that we don't know will ever exist. So I think, until someone gives us a use case that suggests that we should generalise, we might as well keep things simple.

src/backends/perf_pt/decode.c
ltratt: Maybe just two `if`s then.

src/backends/perf_pt/decode.c
vext01: I don't think we need to be quite that verbose, but I agree.

src/backends/perf_pt/decode.c
vext01: Fair enough. FWIW, we have #49 to remind us that this is on the cards.

ltratt: I'm not sure having that as an issue is very useful (*we're* never going to do it AFAICS), but up to you.

src/backends/perf_pt/test_helpers.c
vext01: I think in this case it would be best to leave them apart for the reason above.

src/backends/perf_pt/decode.c
vext01: I think the correct thing to do is to update the README and clearly state that the library is currently for tracing one's self.

src/backends/perf_pt/test_helpers.c
ltratt: OK.

src/backends/perf_pt/decode.c
ltratt: I agree.

examples/simple_example.rs
vext01: Is there a better way than using `last()` to consume the iterator?

vext01: For context:
```
warning: unused `std::iter::Map` which must be used: iterator adaptors are lazy and do nothing unless consumed
  --> examples/simple_example.rs:73:5
   |
73 | /     trace.iter_blocks().take(qty)
74 | |                        .enumerate()
75 | |                        .map(|(i, blk)| println!("  block {}: 0x{:x}", i, blk.unwrap().start_vaddr()));
   | |______________________________________________________________________________________________________^
   |
```

src/backends/perf_pt/decode.c
vext01: Hrm. Actually, I think that's clear from the source. Look at the sole call site of this function:
https://github.com/softdevteam/hwtracer/blob/ffa6ee0a521aa5f201d446a2c4aff4a979b3e3cb/src/backends/perf_pt/mod.rs#L87-L99

examples/simple_example.rs
ltratt: Good point. This should probably be a `for` loops in retrospect:

```rust
for x in trace.iter_blocks().take(qty) {
  // do something with x
}
```

vext01: Using `.last()` works fine. I think with a comment I actually prefer what you were proposing.

```diff
diff --git a/examples/simple_example.rs b/examples/simple_example.rs
index 2260d51..6b083a9 100644
--- a/examples/simple_example.rs
+++ b/examples/simple_example.rs
@@ -69,28 +69,24 @@ fn tracer() -> Box<Tracer> {
 fn print_trace(trace: &Box<Trace>, name: &str, result: u32, qty: usize) {
     let count = trace.iter_blocks().count();
     println!("{}: num_blocks={}, result={}", name, count, result);
-    let mut itr = trace.iter_blocks();
-    let mut end = false;
-    for i in 0..qty {
-        if let Some(blk) = itr.next() {
-            println!("  block {}:\t0x{:x}", i, blk.unwrap().start_vaddr());
-        } else {
-            end = true;
-        }
-    }
-    if !end {
+
+    trace.iter_blocks().take(qty)
+                       .enumerate()
+                       .map(|(i, blk)| println!("  block {}: 0x{:x}", i, blk.unwrap().start_vaddr()))
+                       .last();
+    if count > qty {
         println!("  ... {} more", count - qty);
     }
     println!("");
 }
```

ltratt: The Rust docs say (and I was being an idiot for not thinking of this) "The idiomatic way to write a map for its side effects is to use a for loop instead".

vext01: Fair enough. I'll see what I can come up with.

src/backends/perf_pt/decode.c
vext01: It didn't look too bad in the end.

src/backends/perf_pt/decode.c
vext01: Just ha d a look. It's currently a mix.

vext01:
The last commit addresses your comments.

I think I'd like to rename the functions in `decode.c` so as to make clear we are decoding blocks (not instructions, which libipt supports too). OK by you?

vext01:
Like this ^

examples/simple_example.rs
```diff
@@ -69,28 +69,23 @@ fn tracer() -> Box<Tracer> {
...
-    let mut itr = trace.iter_blocks();
-    let mut end = false;
-    for i in 0..qty {
-        if let Some(blk) = itr.next() {
-            println!("  block {}:\t0x{:x}", i, blk.unwrap().start_vaddr());
-        } else {
-            end = true;
-        }
+
+    for (i, blk) in trace.iter_blocks().take(qty) .enumerate() {
```
ltratt: Stray space before `.enumerate`.

src/backends/perf_pt/decode.c
```diff
@@ -106,11 +105,9 @@ perf_pt_start_decode(void *buf, uint64_t len, int vdso_fd, char *vdso_filename,
     }
 
     // Work around CPU bugs.
-    if (config.cpu.vendor) {
-        if (pt_cpu_errata(&config.errata, &config.cpu) < 0) {
-            failing = true;
-            goto clean;
-        }
+    if ((config.cpu.vendor) && (pt_cpu_errata(&config.errata, &config.cpu) < 0)) {
```
ltratt: No need for brackets around `config.cpu.vendor`.

ltratt:
>I think I'd like to rename the functions in decode.c so as to make clear we are decoding blocks (not instructions, which libipt supports too). OK by you?

Fine with me.

vext01:
So is it squash time?

ltratt:
Yep!

vext01:
splat!

PR #52, opened 2018-03-22
Use a macro to skip tests if the CPU doesn't talk PT.

vext01:
Simplifies the tests a fair amount. Looks good?

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -341,11 +341,7 @@ impl PerfPTTracer {
                 : "eax", "ecx", "edx"
                 : "volatile");
         }
-
-        if ebx_out & (EBX_BIT) != 0 {
-            return true;
-        }
-        false
+        ebx_out & EBX_BIT != 0
     }
 
     fn err_if_destroyed(&self) -> Result<(), HWTracerError> {


ltratt:
I must admit, I don't like this a great deal :/ It introduces a very unobvious control flow that I fear will trip us up one day and doesn't really simplify things enough for that to seem worthwhile.

What you could perhaps do is introduce a `cfg` option that allows you to write things like:

```rust
#[test]
#[cfg(hwtrace_available)]
fn test_blah() { ... }
``

**Comment by vext01** [comment]
Is there a good test framework for Rust maybe? I can't help but think the one we are using is quite limited.

**Comment by ltratt** [comment]
It is *the* Rust framework. I don't know if there are others. But, in this case, I don't see what's limited about it? We want to do conditional compilation, it seems to me, and there seems to be a way to do it?

**Comment by vext01** [comment]
Heh, my last comment came over a little grumpier than intended.

I think I've been spoilt by `py.test`, but the things I miss are:

 * Proper setup/tear down -- There are parts of our test suite which leave temp files behind if the test fails with a panic.
 * Conditionally skipping tests (at runtime), and the notion of "skipping" a test in general.
 * Fixtures

There's a few test frameworks around, but the only one that comes close to my expectations is [Galvanic](https://github.com/mindsbackyard/galvanic-test), which supports setup/tear down and fixtures, but not runtime conditional skipping.

I think we can get by by implementing a macro like you said above (I may try galvanic for my personal projects).

**Comment by ltratt** [comment]
I don't mind trying a different testing framework in this repo.

**Comment by vext01** [comment]
OK. We still need the macro, so I'll do that as part of this PR, then I might try translating the tests to galvanic another time.

**Comment by ltratt** [comment]
Which macro?

**Comment by vext01** [comment]
The one you suggested above do skip the test if the CPU doesn't support PT. I think I can make a macro which embeds the test body into a closure only executed if an outer PT check passes.

**Comment by vext01** [comment]
Wait, your skip there is a compile time check. Is that what we want? I envisaged this as a runtime check.

**Comment by ltratt** [comment]
I was suggesting a compile-time switch. I'm not suggesting that's perfect, but it is what I was suggesting ;)

**Comment by vext01** [comment]
Leave it with me, I think we can do better ;)

**Comment by vext01** [comment]
Hrmph. To write the decorator I had in mind, I'd need to write a syntax extension which disappointingly requires access to nightly features (I'd need the `plugin_registrar`, `quote` and `rustc_private` features enabled) . I'm not sure that's the kind of thing we want to get involved with.

Although a compile time switch is not perfect, I think we can roll with it for now. It's not like we will ever compile the tests on one system and distribute them to another.

**Comment by ltratt** [comment]
Agreed.

**Comment by vext01** [comment]
How about this @ltratt ?

File: build.rs [hunk 1]
```diff
@@ -99,6 +101,32 @@ fn fetch_libipt() {
     env::set_current_dir(&Path::new("..")).unwrap();
 }
 
+// Checks if the CPU supports Intel Processor Trace.
+// We use this to decide whether to run the perf_pt backend tests. Although this would be better as
+// a runtime check, it's OK since we won't distribute the test binary.
+fn cpu_supports_pt() -> bool {
```
vext01: Note that I had to copy this function from the main sources.

build.rs
```diff
@@ -99,6 +101,32 @@ fn fetch_libipt() {
     env::set_current_dir(&Path::new("..")).unwrap();
 }
 
+// Checks if the CPU supports Intel Processor Trace.
+// We use this to decide whether to run the perf_pt backend tests. Although this would be better as
+// a runtime check, it's OK since we won't distribute the test binary.
```
ltratt: I don't think this could ever sensibly be a runtime check (after all, we'll statically compile in the backend!), so we probably delete this comment.

build.rs
```diff
@@ -99,6 +101,32 @@ fn fetch_libipt() {
...
+              mov $1, %eax;
+              mov $2, %ecx;
+              cpuid;"
+            : "={ebx}" (ebx_out)
+            : "i" (LEAF), "i" (SUBPAGE)
+            : "eax", "ecx", "edx"
+            : "volatile");
+    }
+
+    if ebx_out & (EBX_BIT) != 0 {
```
ltratt: Simplify to:

```rust
ebx_out & EBX_BIT != 0
```

ltratt:
A couple of minor comments but basically: looks good!

build.rs
vext01: That's not right. We could build the our code on a system with a CPU with no PT support and distribute it to a system with a CPU that does support PT. Otherwise I would have re-used the existing `perf_pt` compile time flag.

build.rs
vext01: Not sure why that's like that. I will update the source of the paste too.

build.rs
ltratt: Dumb question: why can't we use the `perf_pt` flag?

vext01: That decides if we are going to build with the `perf_pt` backend. The prerequisite is that the build system has a new enough Linux Perf API. It says nothing about whether the CPU on the build system supports PT or not.

ltratt: Is it ever useful to say "perf is too old, but we're going to do stuff with PT support anyway"?

vext01: If perf is too old the PT backend cannot be built and the tests cannot be run.

ltratt: Yes, I understand that, but it doesn't quite answer my question. Basically, will we ever make use of a situation where `perf_pt == false` but this function you've added returns true?

vext01: Nope. But we do need a separate flag for the case where Perf is new enough, but the CPU on the build machine doesn't have PT. Do you follow?

ltratt: Ah, yes. I wonder if we could just collapse that into one flag though, since we don't *really* care about the individual parts?

vext01: We could collapse it into one, but it would mean that we can't distribute binaries if the build system has no PT on the chip.

ltratt: Aha, OK, you've convinced me!

vext01:
OK. I've pushed the `EBX` change to the main sources separately as it's not really related. OK to squash?

vext01:
Ah, i see!

ltratt:
Now we can squash :)

vext01:
splat.

PR #53, opened 2018-03-28
Handle (some) PT event packets.

### LINKED ISSUE #4: Check for packet loss

vext01:
This change makes hwtracer look at event packets in the packet stream.

 * Most of the cases remains as "ignore this packet".
 * The overflow packet crashes loudly for now (until we have proper error handling).
 * Any case which I am not sure about is caught by an assertion in the default arm of the switch.

For each case I've added a comment saying what the packet does so that we can see at a glance without wading through the manual.

Fixes #4.

### DIFF
--- a/src/backends/perf_pt/decode.c
+++ b/src/backends/perf_pt/decode.c
@@ -58,7 +58,7 @@ struct load_self_image_args {
 };
 
 // Private prototypes.
-static int drain_events(struct pt_block_decoder *, int);
+static bool handle_events(struct pt_block_decoder *, int *);
 static bool load_self_image(struct load_self_image_args *);
 static int load_self_image_cb(struct dl_phdr_info *, size_t, void *);
 
@@ -169,8 +169,10 @@ clean:
 bool
 perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
                    uint64_t *addr) {
-    // If there are events pending, drain them first.
-    *decoder_status = drain_events(decoder, *decoder_status);
+    // If there are events pending, look at those first.
+    if (handle_events(decoder, decoder_status) != true) {
+        return false;
+    }
     if (*decoder_status < 0) {
         // Error.
         return false;
@@ -185,7 +187,7 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
     struct pt_block block;
     *decoder_status = pt_blk_next(decoder, &block, sizeof(block));
     // Other +ve decoder status codes can arise here. We ignore them for now,
-    // and let them be detected by drain_events() above when we are next
+    // and let them be detected by handle_events() above when we are next
     // called.
     if (*decoder_status < 0) {
         if (*decoder_status == -pte_eos) {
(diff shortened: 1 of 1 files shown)


src/backends/perf_pt/decode.c
```diff
@@ -169,8 +169,10 @@ perf_pt_init_block_decoder(void *buf, uint64_t len, int vdso_fd, char *vdso_file
 bool
 perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
                    uint64_t *addr) {
-    // If there are events pending, drain them first.
-    *decoder_status = drain_events(decoder, *decoder_status);
+    // If there are events pending, look at those first.
+    if (handle_events(decoder, decoder_status) != true) {
```
ltratt: This seems like a slightly odd reason for returning `false`?

ltratt: Perhaps this should be an `assert`?

src/backends/perf_pt/decode.c
```diff
@@ -213,21 +215,94 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
         struct pt_event event;
-        decoder_status = pt_blk_event(decoder, &event, sizeof(event));
-        if (decoder_status < 0) {
-            return decoder_status;
+        *decoder_status = pt_blk_event(decoder, &event, sizeof(event));
+        if (*decoder_status < 0) {
+            return *decoder_status;
+        }
+
+        switch (event.type) {
```
ltratt: Something's wrong with the indentation of the `case`: is it a tabs issue? They should be indented, right?

src/backends/perf_pt/decode.c
vext01: All of the cases I don't expect to see are already caught by an assert in the callee. Returning false here only applies to buffer overflow or `pt_blk_event()` failing.

src/backends/perf_pt/decode.c
vext01: It was actually intentional. I've seen all kinds of indent for switch statements. This one saves vertical space since we have large block comments. I can indent it once more if you feel strongly.

vext01:
I've also seen an occasional assertion failure for packet type 2, so I need to add a handler for that. Watch this space.

vext01:
Simple fix to handle packet type 2.

Let me know about the other bits.

vext01:
The last commit indents the switch and re-wraps the comments. Up to you if we keep it or not.

vext01:
It only adds one line, so I suggest we keep it.

src/backends/perf_pt/decode.c
ltratt: I'm not saying it's a bad idea, but it is unusual, so I'm inclined to follow convention.

vext01:
So I think we are ready to squash? I have one more PR waiting on this one.

ltratt:
Please squash.

vext01:
Splat.

PR #54, opened 2018-03-29
Errors Part 1

vext01:
This refactors the error handling as we discussed the other day.

I'm only dealing with Rust-level errors at this time, C-level errors will follow. See the `XXX`.

I'm still questioning weather the "fatal" / "non-fatal" distinction is useful. Is "common to all backends" / "not common to all backends" a better distinction?

Thoughts?

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -47,7 +47,6 @@ pub enum HWTracerError {
     NumParseInt(num::ParseIntError),
     // Our own errors.
     CFailure,
-    ElfError(String),
     HardwareSupport(String),
     InvalidFileName(String),
     TracerAlreadyStarted,
@@ -89,7 +88,6 @@ impl Display for HWTracerError {
             HWTracerError::NumParseInt(ref e) => write!(f, "{}", e),
             HWTracerError::HardwareSupport(ref m) => write!(f, "Hardware support: {}", m),
             HWTracerError::CFailure => write!(f, "Calling to C failed"),
-            HWTracerError::ElfError(ref m) => write!(f, "ELF error: {}", m),
             HWTracerError::InvalidFileName(ref n) => write!(f, "Invalid file name: `{}'", n),
             HWTracerError::TracerAlreadyStarted => write!(f, "Tracer already started"),
             HWTracerError::TracerDestroyed => write!(f, "Tracer destroyed"),
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -35,64 +35,52 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use std::{io, ffi, num};
 use std::fmt::{self, Formatter, Display};
+use std::error::Error;
+use TracerState;
 
 #[derive(Debug)]
 pub enum HWTracerError {
-    // Wrapped errors from elsewhere.
-    FFIIntoString(ffi::IntoStringError),
-    FFINul(ffi::NulError),
-    IO(io::Error),
-    NumParseInt(num::ParseIntError),
-    // Our own errors.
-    CFailure,
-    HardwareSupport(String),
-    InvalidFileName(String),
-    TracerAlreadyStarted,
-    TracerDestroyed,
-    TracerNotStarted,
-    TracingNotPermitted(String),
-}
-
-impl From<ffi::IntoStringError> for HWTracerError {
-    fn from(err: ffi::IntoStringError) -> Self {
-        HWTracerError::FFIIntoString(err)
-    }
-}
-
-impl From<ffi::NulError> for HWTracerError {
-    fn from(err: ffi::NulError) -> Self {
-        HWTracerError::FFINul(err)
-    }
-}
-
-impl From<io::Error> for HWTracerError {
-    fn from(err: io::Error) -> Self {
-        HWTracerError::IO(err)
-    }
-}
-
-impl From<num::ParseIntError> for HWTracerError {
-    fn from(err: num::ParseIntError) -> Self {
-        HWTracerError::NumParseInt(err)
-    }
+    HWBufferOverflow,         // The trace buffer being used by the hardware overflowed.
+                              // This is considered a non-fatal error since retrying the tracing
+                              // may succeed.
+    NoHWSupport(String),      // The hardware doesn't support a required feature. Not fatal for the
+                              // same reason as `Permissions`. This may be non-fatal depending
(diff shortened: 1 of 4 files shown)


src/errors.rs
```diff
@@ -35,64 +35,55 @@
...
-        HWTracerError::FFIIntoString(err)
-    }
-}
+    // Non-fatal errors.
+    HWBufferOverflow,     // The trace buffer being used by the hardware overflowed.
+                          // This is considered a non-fatal error since retrying the tracing may
+                          // succeed.
+    HWSupport(String),    // The hardware doesn't support a required feature. Not fatal for the
+                          // same reason as `TracingPerms`.
+    TracingPerms(String), // Tracing is not permitted. Non-fatal as the consumer could carry on
```
ltratt: Not a big fan of this name: `PermissionDenied` or something might be clearer?

src/errors.rs
```diff
@@ -35,64 +35,55 @@
...
-impl From<ffi::IntoStringError> for HWTracerError {
-    fn from(err: ffi::IntoStringError) -> Self {
-        HWTracerError::FFIIntoString(err)
-    }
-}
+    // Non-fatal errors.
+    HWBufferOverflow,     // The trace buffer being used by the hardware overflowed.
+                          // This is considered a non-fatal error since retrying the tracing may
+                          // succeed.
+    HWSupport(String),    // The hardware doesn't support a required feature. Not fatal for the
```
ltratt: This sounds like it's a positive, but it's a negative, so I think we want something like `NotHWSupported`?

src/errors.rs
```diff
@@ -35,64 +35,55 @@
...
+                          // without tracing support.
 
-impl From<ffi::NulError> for HWTracerError {
-    fn from(err: ffi::NulError) -> Self {
-        HWTracerError::FFINul(err)
-    }
+    // Fatal errors.
+    CFailure,                 // Something went wrong in C code.
+                              // ^ XXX will be replaced with an errno mechanism.
+    TracerState(TracerState), // The tracer is in the wrong state to do the requested task.
```
ltratt: I wonder whether this should just be an internal `panic!`?

src/errors.rs
vext01: Do you intend to make it a more general error? Otherwise we should qualify it with something that makes it's specific to tracing capabilities. e.g. `TraceingPermissionsDenied`, which I find too long.

src/errors.rs
vext01: How about `NoHWSupport`?

src/errors.rs
ltratt: Since everything here relates to tracing, I don't see how that prefix is useful TBH.

src/errors.rs
vext01: The user can easily trigger this, so I think not.

```
tracer.start_tracing();
tracer.start_tracing(); // boom
```

ltratt:
> I'm still questioning weather the "fatal" / "non-fatal" distinction is useful. Is "common to all backends" / "not common to all backends" a better distinction?

I don't think these two ideas are mutually exclusive. If we make errors common to backends, there must be an implicit guarantee that "error E means something can be retried but error F means something can't be retried as the system has gone horribly wrong."

src/errors.rs
ltratt: Right, but since everyone will just do:

```rust
tracer.start_tracing().unwrap()
```

it seems to me that all we're doing is forcing the user to do `.unwrap()` everywhere (they're never *not* going to do `unwrap()`, as they can't recover from this error)?

src/errors.rs
ltratt: Works for me.

vext01:
I think we've actually made the same observation. Any of these errors could be retryable or not depending upon the scenario.

Should we have a top level error type which is a `Fatal`/`Retryable` and then inside in an inner error type?

vext01:
I'm fine with `Fatal` / `Nonfatal`, but if it's too heavy for your taste, we have to think again.

ltratt:
I think it's worth trying. If it feels too icky, we can go back.

vext01:
OK

src/errors.rs
```diff
@@ -35,66 +35,55 @@
...
+    FatalError(Box<Error>),   // All other errors. Boxed so that individual backends can choose
```
ltratt: Should we call this `UnknownError` or `CustomError` or similar?

vext01: `io::Error` in the Rust std lib uses both `Other` and `Custom`. It's a bit more complicated there, as they separate the error type from it's representation, but it all boils down to:

```
enum Repr {
    Os(i32),
    Simple(ErrorKind),
    Custom(Box<Custom>),
}
```

Which is amusingly recursive.

https://doc.rust-lang.org/src/std/io/error.rs.html#77

The last push uses `Custom` for our "other error type".

vext01:
How is this?

src/errors.rs
vext01: How would you feel if a library you were using as a black box crashed with a panic instead of giving you a meaningful error?

src/errors.rs
vext01: OK, let's go for `Permissions`.

src/errors.rs
ltratt: A `panic` can give a useful error message. In this case, the onus is on the user to set up their sequence of calls to do the right thing, and it seems unlikely that they would ever do anything other than `unwrap`. Anyway, I don't feel strongly about it, so feel free to ignore me on this one.

vext01: Another part of me also feels that I'm OK with the user crashing their own program with a hasty `unwrap()` as that clearly sends a message that the user did something wrong (starting a tracer twice). It feels somewhat rude for us to crash the user's program with something so easily triggered.

vext01:
I lost your comment with the last push, so here it is pasted from MM:

> A panic can give a useful error message. In this case, the onus is on the user to set up their sequence of calls to do the right thing, and it seems unlikely that they would ever do anything other than unwrap. Anyway, I don't feel strongly about it, so feel free to ignore me on this one.

I guess it's unlikely that anyone would see this other than during development, so maybe a panic is fine. On the other hand, removing this error type would not change the signature of (e.g.) `start_tracing` as it can return other errors too. For this reason, we might as well keep the error type. Agree?

ltratt:
Agreed.

vext01:
OK, so if you are happy with the previous tweaks, it's time to squash I think.

ltratt:
Agreed.

vext01:
splat.

PR #56, opened 2018-04-04
Proper C Error Handling.

vext01:
This implements a error handling scheme for the perf_pt backend which is capable of communicating libipt and errno style errors.

Instead of just returning "something went wrong", we now set a type and code using `perf_pt_set_err`, which Rust then converts into a high-level error type.

It took a little longer than expected due to some fiddly cases.

Notes:

 * First commit is an unrelated OCD fix.
 * The overflow case is now properly handled, rather than being munged together with all other errors. So now we can know when to retry.
 * Removed an error indicator from a function which cannot fail  (`read_aux()`).
 * Fixed a NULL pointer dereference (I'll point this out).
 * Added a field into the tracer context to store errors that can occur inside the thread. It needs to live here so that the error struct can live for the entirety of the tracing session. Then in `stop_tracing` we check if anything went wrong "while we were away".

I did wonder if I need to add a comment to every function accepting an error struct. What do you think? Overkill?

Thanks

Fixes #56

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -78,6 +78,8 @@
  */
 struct tracer_ctx {
     pthread_t           tracer_thread;      // Tracer thread handle.
+    struct perf_pt_cerror
+                        tracer_thread_err;  // Errors from inside the tracer thread.
     int                 stop_fds[2];        // Pipe used to stop the poll loop.
     int                 perf_fd;            // FD used to talk to the perf API.
     void                *aux_buf;           // Ptr to the start of the the AUX buffer.
@@ -119,31 +121,33 @@ struct tracer_thread_args {
     void                *aux_buf;           // The AUX buffer itself;
     struct perf_event_mmap_page
                         *base_header;       // Pointer to the header in the base buffer.
+    struct perf_pt_cerror
+                        *err;               // Errors generated inside the thread.
 };
 
 
 // Private prototypes.
-static bool read_aux(void *, struct perf_event_mmap_page *, struct perf_pt_trace *);
+static void read_aux(void *, struct perf_event_mmap_page *,
+                     struct perf_pt_trace *);
 static bool poll_loop(int, int, struct perf_event_mmap_page *, void *,
-                      struct perf_pt_trace *);
+                      struct perf_pt_trace *, struct perf_pt_cerror *);
 static void *tracer_thread(void *);
-static int open_perf(pid_t target_tid);
+static int open_perf(pid_t target_tid, struct perf_pt_cerror *);
 
 // Exposed Prototypes.
-struct tracer_ctx *perf_pt_init_tracer(struct tracer_conf *);
-bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *);
-bool perf_pt_stop_tracer(struct tracer_ctx *tr_ctx);
-bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx);
+struct tracer_ctx *perf_pt_init_tracer(struct tracer_conf *, struct perf_pt_cerror *);
+bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct perf_pt_cerror *);
+bool perf_pt_stop_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
+bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
 
 /*
  * Read data out of the AUX buffer.
  *
  * Reads from `aux_buf` (whose meta-data is in `hdr`) into `trace`.
- *
- * Returns true on success and false otherwise.
  */
-static bool
-read_aux(void *aux_buf, struct perf_event_mmap_page *hdr, struct perf_pt_trace *trace)
+void
+read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
+         struct perf_pt_trace *trace)
 {
     // We use atomic loads and stores with orderings to avoid concurrency
     // issues on the AUX state mutably shared between us and the kernel.
@@ -179,7 +183,6 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr, struct perf_pt_trace *
     }
     trace->len += new_data_size;
     atomic_store_explicit((_Atomic __u64 *) &hdr->aux_tail, head, memory_order_release);
-    return true;
 }
 
 /*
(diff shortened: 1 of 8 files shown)


src/backends/perf_pt/collect.c
```diff
@@ -401,13 +415,14 @@ perf_pt_init_tracer(struct tracer_conf *tr_conf)
...
+        perf_pt_set_err(err, perf_pt_cerror_errno, errno);
         failing = true;
         goto clean;
     }
 
 clean:
-    if (failing) {
-        perf_pt_free_tracer(tr_ctx);
+    if ((failing) && (tr_ctx != NULL)) {
+        perf_pt_free_tracer(tr_ctx, err);
```
vext01: Here's the NULL pointer dereference fix.

src/backends/perf_pt/perf_pt_private.h
```diff
@@ -47,6 +47,18 @@
...
+enum perf_pt_cerror_type {
+    perf_pt_cerror_unknown,
+    perf_pt_cerror_errno,
+    perf_pt_cerror_ipt,
+};
+
+struct perf_pt_cerror {
+    int typ;    // Is it a errno or libipt error code.
+    int code;   // The error code itself.
+    bool set;   // Used to ensure newer failures don't overwrite older ones.
```
ltratt: The `bool` seems pointless (and bloats the struct because of alignment!): if you have an entry in the enum `Unused` then it subsumes the bool AFAICS?

vext01: Good idea.

vext01: Fix pushed.

src/backends/perf_pt/collect.c
ltratt: There's no need for the brackets around `(failing)` or `(tr_ctx != NULL)`.

src/backends/perf_pt/collect.c
```diff
@@ -466,7 +466,6 @@ perf_pt_start_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_trace *trace, str
     // coming from inside the thread. We initialise it to "no errors".
     tr_ctx->tracer_thread_err.typ = 0;
```
ltratt: I think this should be `type = PerfPTCErrorType.Unused` or similar?

src/backends/perf_pt/mod.rs
```diff
@@ -80,7 +80,9 @@ impl Error for LibIPTError {
 
 #[repr(C)]
 #[allow(dead_code)] // Only C constructs these.
+#[derive(PartialEq)]
 enum PerfPTCErrorType {
```
ltratt: If we want to make this Rust-y it might be better to use `Kind` rather than `Type`.

src/backends/perf_pt/collect.c
vext01: Ah, yes, it should be `perf_pt_cerror_unused` so as to reference the enum.

src/backends/perf_pt/decode.c
```diff
@@ -315,8 +326,16 @@ handle_events(struct pt_block_decoder *decoder, int *decoder_status) {
 static bool
 load_self_image(struct load_self_image_args *args)
 {
-    return ((dl_iterate_phdr(load_self_image_cb, args) == 0) &&
-            (fsync(args->vdso_fd) == 0));
+    if (dl_iterate_phdr(load_self_image_cb, args) != 0) {
```
ltratt: Shouldn't this set `perf_pt_set_err`?

src/backends/perf_pt/mod.rs
vext01: Sure, are you happy for me to update the name on the C side too?

ltratt: Yep.

src/backends/perf_pt/decode.c
vext01: The error struct is inside `args` and the callback sets the error directly. We are limited by the API here.

ltratt: We can still be clear that *some* sort of error happened though, can't we?

vext01: The error struct is an alias to the one used in the parent, so that errors are properly propagated.

src/backends/perf_pt/decode.c
```diff
@@ -124,24 +133,28 @@ perf_pt_init_block_decoder(void *buf, uint64_t len, int vdso_fd, char *vdso_file
...
     // Build and load a memory image from which to recover control flow.
     struct pt_image *image = pt_image_alloc(NULL);
     if (image == NULL) {
+        perf_pt_set_err(err, perf_pt_cerror_unknown, 0);
         failing = true;
         goto clean;
     }
 
-    struct load_self_image_args load_args = {image, vdso_fd, vdso_filename};
+    struct load_self_image_args load_args = {image, vdso_fd, vdso_filename, err};
```
vext01: Here is where the alias is created.

src/backends/perf_pt/collect.c
vext01: I agree with `failing` but the latter condition is easier to read with parens, isn't it? The whole code-base uses that style.

vext01:
Pushed a commit to address your comments so far. You were right that I can use an enum as a type, so I've made that change too.

src/backends/perf_pt/perf_pt_private.h
```diff
@@ -47,6 +47,18 @@
...
 
+enum perf_pt_cerror_kind {
+    perf_pt_cerror_unused,
+    perf_pt_cerror_unknown,
+    perf_pt_cerror_errno,
+    perf_pt_cerror_ipt,
+};
+
+struct perf_pt_cerror {
+    enum perf_pt_cerror_kind kind; // What sort of error is this?
```
vext01: So here, we can use the enum type rather than int :)

src/backends/perf_pt/decode.c
ltratt: OK.

src/backends/perf_pt/collect.c
ltratt: It's a matter of taste, I admit, but since I'm not a Lisper, I rarely find that more brackets help. But if that's the style of this file, we might as well stick with it.

ltratt:
I think we're ready to squash?

vext01:
I think so.

ltratt:
Then squash away!

vext01:
Splat.

ltratt:
I don't mind it being 1 commit, but I think you've got the wrong commit title!

vext01:
Oops. That wasn't my intention at all!

vext01:
Try this. 2 commits.

PR #58, opened 2018-04-18
Various buffer related improvements.

vext01:
Several buffer-related fixes and improvements.

It's probably best to look at individual commits.

(I also improved some comments to help us next time we are in this code)

I've not included one of my tests which uncovered a potential bug, as I'm working on that with Markus in https://github.com/01org/processor-trace/issues/39 . I'll put that in another PR when I know what's gone wrong.

OK?

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -226,6 +226,8 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
             goto done;
         }
 
+        // POLLIN on pfds[0]: Overflow event on either the Perf AUX or data buffer.
+        // POLLHUP on pfds[1]: Tracer stopped by parent.
         if ((pfds[0].revents & POLLIN) || (pfds[1].revents & POLLHUP)) {
             read_aux(aux, mmap_hdr, trace, err);
 
@@ -234,6 +236,7 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
             }
         }
 
+        // The traced thread exited.
         if (pfds[0].revents & POLLHUP) {
             break;
         }
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -68,6 +68,8 @@
 #define MAX_OPEN_PERF_TRIES  5000
 #define OPEN_PERF_WAIT_NSECS 1000 * 20
 
+#define AUX_BUF_WAKE_RATIO 0.5
+
 #ifndef INFTIM
 #define INFTIM -1
 #endif
@@ -132,7 +134,7 @@ static bool read_aux(void *, struct perf_event_mmap_page *,
 static bool poll_loop(int, int, struct perf_event_mmap_page *, void *,
                       struct perf_pt_trace *, struct perf_pt_cerror *);
 static void *tracer_thread(void *);
-static int open_perf(pid_t target_tid, struct perf_pt_cerror *);
+static int open_perf(pid_t, size_t, struct perf_pt_cerror *);
 
 // Exposed Prototypes.
 struct tracer_ctx *perf_pt_init_tracer(struct tracer_conf *, struct perf_pt_cerror *);
@@ -247,7 +249,7 @@ done:
  * Returns a file descriptor, or -1 on error.
  */
 static int
-open_perf(pid_t target_tid, struct perf_pt_cerror *err) {
+open_perf(pid_t target_tid, size_t aux_bufsize, struct perf_pt_cerror *err) {
     struct perf_event_attr attr;
     memset(&attr, 0, sizeof(attr));
     attr.size = sizeof(attr);
@@ -281,6 +283,10 @@ open_perf(pid_t target_tid, struct perf_pt_cerror *err) {
 
     // No skid.
     attr.precise_ip = 3;
+
+    attr.watermark = 0;
+    attr.wakeup_events = 1;
+    attr.aux_watermark = (size_t) ((double) aux_bufsize * getpagesize()) * AUX_BUF_WAKE_RATIO;
 
     // Acquire file descriptor through which to talk to Intel PT. This syscall
     // could return EBUSY, meaning another process or thread has locked the
@@ -381,7 +387,7 @@ perf_pt_init_tracer(struct tracer_conf *tr_conf, struct perf_pt_cerror *err)
     tr_ctx->perf_fd = -1;
 
     // Obtain a file descriptor through which to speak to perf.
-    tr_ctx->perf_fd = open_perf(tr_conf->target_tid, err);
+    tr_ctx->perf_fd = open_perf(tr_conf->target_tid, tr_conf->aux_bufsize, err);
     if (tr_ctx->perf_fd == -1) {
         perf_pt_set_err(err, perf_pt_cerror_errno, errno);
         failing = true;
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -127,8 +127,8 @@ struct tracer_thread_args {
 
 
 // Private prototypes.
-static void read_aux(void *, struct perf_event_mmap_page *,
-                     struct perf_pt_trace *);
+static bool read_aux(void *, struct perf_event_mmap_page *,
+                     struct perf_pt_trace *, struct perf_pt_cerror *);
 static bool poll_loop(int, int, struct perf_event_mmap_page *, void *,
                       struct perf_pt_trace *, struct perf_pt_cerror *);
 static void *tracer_thread(void *);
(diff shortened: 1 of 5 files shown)


vext01:
Hold off on this for the moment. `--sb:warn` won't do anything and can be removed.

vext01:
Ready for review.

src/backends/perf_pt/collect.c
```diff
@@ -171,8 +171,23 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
...
 
-    // For now we simply crash out if we exhaust the buffer.
-    assert(trace->len + new_data_size <= trace->capacity);
+    // Decide if we have space for the new data, and reallocate the trace
+    // storage buffer if more space is required.
+    __u64 required_capacity = trace->len + new_data_size;
+    if (required_capacity > trace->capacity) {
+        // Over-allocate to 2x what we need, checking that the result fits in
+        // the size_t argument of realloc(3).
+        assert(required_capacity < SIZE_MAX / 2);
```
ltratt: What's the point of this? If someone allocates *that* much memory, `realloc` will return `null` (assuming Linux doesn't infinitely overcommit?).

src/backends/perf_pt/collect.c
```diff
@@ -171,8 +171,23 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
         new_data_size = (size - tail) + head;
     }
 
-    // For now we simply crash out if we exhaust the buffer.
-    assert(trace->len + new_data_size <= trace->capacity);
+    // Decide if we have space for the new data, and reallocate the trace
```
ltratt: I think this comment is a bit on the obvious side. If we keep it, let's at least chop it down to 1 line.

src/backends/perf_pt/collect.c
```diff
@@ -283,6 +285,11 @@ open_perf(pid_t target_tid, struct perf_pt_cerror *err) {
     // No skid.
     attr.precise_ip = 3;
 
+    // Wake up each time the AUX buffer becomes AUX_BUF_WAKE_PERCENT% full.
+    attr.watermark = 0;
+    attr.wakeup_events = 1;
+    attr.aux_watermark = (size_t) ((double) aux_bufsize * getpagesize()) * AUX_BUF_WAKE_PERCENT;
```
ltratt: I'm not sure I quite get what this line is doing.

src/backends/perf_pt/collect.c
```diff
@@ -227,6 +227,10 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
             goto done;
         }
 
+        // POLLIN on pfds[0]: Overflow event on either the Perf AUX or data buffer.
+        // POLLHUP on pfds[1]: Tracer stopped by parent.
+        // It's important we read any remaining trace data out of the AUX
```
ltratt: Why is it important?

src/backends/perf_pt/mod.rs
```diff
@@ -406,6 +406,18 @@ impl PerfPTTracer {
             return Err(HWTracerError::NoHWSupport("Intel PT not supported by CPU".into()));
         }
 
+        // Check for inavlid configuration.
+        fn power_of_2(v: size_t) -> bool {
+            !(v <= 0) && ((v & (v - 1)) == 0)
+        }
+        if !power_of_2(config.data_bufsize) {
+            return Err(HWTracerError::BadConfig(String::from("data_bufsize must be a positive power of 2")));
```
ltratt: If we're going to have a `power_of_2` function (which seems reasonable), can't we chuck most of this boilerplate in there too?

src/errors.rs
```diff
@@ -52,6 +52,7 @@ pub enum HWTracerError {
     Permissions(String),      // Tracing is not permitted using this backend.
     Errno(c_int),             // Something went wrong in C code.
     TracerState(TracerState), // The tracer is in the wrong state to do the requested task.
+    BadConfig(String),        // The tracer configuration was invalid in some way.
```
ltratt: Drop "in some way".

ltratt:
Various minor comments, although only 1 is even vaguely substantial.

src/backends/perf_pt/collect.c
vext01: The assertion is protecting us against a heap overflow.

`realloc` accepts a `size_t` for the new size, and we are about to multiply a (somewhat) user controlled number by 2 before reallocating that much.

Consider the case where `required_capacity` is (for example) `SIZE_MAX / 2 + 1`.  When we multiply that by 2, it overflows to 2. When we pass that to `realloc` we would end up shrinking the buffer (not growing it) leaving us open to overflowing the buffer later.

src/backends/perf_pt/collect.c
vext01: It's to do with the  Perf when to notifies us that the buffer is getting full.

From the `perf_event_open` manual:
```
       aux_watermark (since Linux 4.1)
              This specifies how much data is required to trigger a PERF_RECORD_AUX sample.
```

This means we wake up the poll loop when the buffer is `AUX_BUF_WAKE_PERCENT` full.

The AUX buffer size must be a power of two pages (our API lets the user configure the number of pages [hence `getpagesize()`], checking it's a power of 2 along the way).

src/backends/perf_pt/collect.c
vext01: Actually, since I raised this PR I've learned more, so we can probably kill this part of the comment.

src/backends/perf_pt/mod.rs
vext01: Sure.

I also considered moving it into `util.rs` as it's something we might need in other places later. In that case we don't want the error handling in there.

Thoughts?

src/backends/perf_pt/collect.c
ltratt: OK, I get the general idea. But if this is user controlled, it should be an error path, not an `assert`, right?

src/backends/perf_pt/collect.c
ltratt: I sort of get that, but I don't get why we `* AUX_BUF_WAKE_PERCENT`. That desn't seem to match the description from the manual?

src/backends/perf_pt/mod.rs
ltratt: I'm fine with having it here  -- if it's not used elsewhere, there's no advantage to moving it.

src/backends/perf_pt/collect.c
vext01: ` aux_bufsize * getpagesize()`  is the byte size of the buffer in total. `AUX_BUF_WAKE_PERCENT` is our own constant define as such:

```
#define AUX_BUF_WAKE_PERCENT 0.25
```

By multiplying by 0.25 we wake up the poll loop when the AUX buffer is roughly 25% full.

Perhaps that macro needs a better name?

[I think we will probably need to make this configurable in the future]

ltratt: A better name might not hurt. 25% also seems a really low value? I'm also not sure how keen I am on the floating point thing here. We might use this on machines where floating point calculations are really slow. Admittedly ` * 0.25` will presumably optimise to `>> 2` but maybe we should make that idea a bit more explicit? Perhaps only allow people to express how many bits to the right we'll shift?

src/backends/perf_pt/collect.c
vext01: Having thought some more, should we perhaps introduce a new error type to indicate that a trace is too long? That's what this effectively means, and I think it's a scenario we should be able to recover from.

If that's what we want, the question then is how to communicate that meaning back to Rust. At current C code can either indicate an errno or a libipt error number.

ltratt: Ah, yes, we will *definitely* encounter this case: if we're unlucky and e.g. trace the last iteration of a loop (in an interpreter that can't catch this), the trace will just go on and on...

As to how to structure it... dunno ;)

src/backends/perf_pt/collect.c
vext01: > We might use this on machines where floating point calculations are really slow.

I wouldn't worry about that. This code only executes on Intel machines new enough to have PT.

> 25% also seems a really low value?

It is. I've not yet decided a suitable default, but I do know that 75% with the default AUX buffer size causes underruns. How about 50%? [Of course a suitable value really depends how big your AUX buffer is.]

ltratt: 50% seems a reasonable starting point.

src/backends/perf_pt/collect.c
vext01: Would you be OK with me adding a new error "kind" for special cases like this?

ltratt: I can't quite picture things well enough to know what's best, so I suggest you go ahead, then we'll work things out afterwards.

vext01: Maybe that's a candidate for a separate PR? Is that OK?

ltratt: Provided it's the next PR, OK. We don't want to allow this one to fester for to long, or it'll get difficult to change.

src/backends/perf_pt/mod.rs
vext01: Wait, we need to raise two different error messages for the two call sites, so I think we probably want to leave this as-is.

vext01:
OK give this a shot.

ltratt:
Works for me. If you squash, I can merge.

vext01:
splat

PR #59, opened 2018-04-19
New error kind and return code for indicating a trace is too long.

vext01:
I think this it roughly what we'd want to replace the assertion that we discussed in the previous PR.

While we are here, I was overloading `pte_overflow` in one place. We can use our new error kind to flag overflow errors and do away with a crusty utility function.

Looks good?

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -178,8 +178,11 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
     if (required_capacity > trace->capacity) {
         // Over-allocate to 2x what we need, checking that the result fits in
         // the size_t argument of realloc(3).
-        // XXX replace assertion with a "trace too long" error kind.
-        assert(required_capacity < SIZE_MAX / 2);
+        if (required_capacity >= SIZE_MAX / 2) {
+            // We would overflow the size_t argument of realloc(3).
+            perf_pt_set_err(err, perf_pt_cerror_errno, ENOMEM);
+            return false;
+        }
         size_t new_capacity = required_capacity * 2;
         void *new_buf = realloc(trace->buf, new_capacity);
         if (new_buf == NULL) {


src/backends/perf_pt/collect.c
```diff
@@ -178,8 +178,11 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
     if (required_capacity > trace->capacity) {
         // Over-allocate to 2x what we need, checking that the result fits in
         // the size_t argument of realloc(3).
-        // XXX replace assertion with a "trace too long" error kind.
-        assert(required_capacity < SIZE_MAX / 2);
+        if (required_capacity >= SIZE_MAX / 2) {
+            // The new size would overflow the size_t argument of realloc(3).
+            perf_pt_set_err(err, perf_pt_cerror_other, PERF_PT_TRACE_TOO_LONG);
```
ltratt: Ok, so I now think this isn't really about the "trace being too long", more "you asked us to allocate terrabytes of memory"?

ltratt: I didn't fully understand what you were getting at before. "Trace too long" is a misleading description IMHO: the trace isn't too long; rather, you're *asking* for the space for the trace to be ridiculously big. Is anyone ever going to hit this on a 64 bit machine? Nope. On a 32 bit machine... *maybe*... but probably not. And if someone was stupid enough to try allocating terrabytes of data, are they they going to be clever enough to reduce their request? Probably not ;) I think my original suggestion of `errx` might be the simplest way out of this.

vext01: For me it's either flag `ENOMEM` or as you prefer (?) `errx`. All I care is that the number isn't allowed to overflow. Modern linux no longer over-commits, so even `realloc` would return `ENOMEM` if there's not enough physical memory.

ltratt: Maybe we should just return `ENOMEM` in either case then?

vext01:
Here we go. OK to squash?

ltratt:
Works for me. Are you happy with it?

vext01:
Fine by me.

ltratt:
OK then, time to squash!

vext01:
Splat.

PR #60, opened 2018-04-24
Read data buf

vext01:
Here is the first wave of bug fixes that I've accumulated over the past week or so.

See individual commits for more info.

### DIFF
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -258,17 +258,14 @@ bool
 read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
          struct perf_pt_trace *trace, struct perf_pt_cerror *err)
 {
-    // We use atomic loads and stores with orderings to avoid concurrency
-    // issues on the AUX state mutably shared between us and the kernel.
-    // Without these, for example the compiler might be able to load data
-    // from `aux_buf` before reading from the `hdr->aux_head`.
+    // Use of atomics here for the same reasons as for handle_sample().
     __u64 head_monotonic =
             atomic_load_explicit((_Atomic __u64 *) &hdr->aux_head,
                                  memory_order_acquire);
-    __u64 size = hdr->aux_size; // No atomic load. Not mutated in-kernel.
+    __u64 size = hdr->aux_size; // No atomic load. Constant value.
     __u64 head = head_monotonic % size; // Head must be manually wrapped.
     __u64 tail = atomic_load_explicit((_Atomic __u64 *) &hdr->aux_tail,
-                                 memory_order_acquire);
+                                 memory_order_relaxed);
 
     // Figure out how much more space we need in the trace storage buffer.
     __u64 new_data_size;
@@ -300,14 +297,16 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
         trace->buf = new_buf;
     }
 
-    // Finally copy the AUX buffer into the trace buffer.
+    // Finally append the new AUX data to the end of the trace storage buffer.
     if (tail <= head) {
-        memcpy(trace->buf, aux_buf + tail, head - tail);
+        memcpy(trace->buf + trace->len, aux_buf + tail, head - tail);
+        trace->len += head - tail;
     } else {
-        memcpy(trace->buf, aux_buf + tail, size - tail);
-        memcpy(trace->buf, aux_buf, head);
+        memcpy(trace->buf + trace->len, aux_buf + tail, size - tail);
+        trace->len += size - tail;
+        memcpy(trace->buf + trace->len, aux_buf, head);
+        trace->len += size + head;
     }
-    trace->len += new_data_size;
     atomic_store_explicit((_Atomic __u64 *) &hdr->aux_tail, head, memory_order_release);
     return true;
 }
--- a/src/backends/perf_pt/collect.c
+++ b/src/backends/perf_pt/collect.c
@@ -59,6 +59,7 @@
 #include <sys/stat.h>
 #include <time.h>
 #include <stdatomic.h>
+#include <intel-pt.h>
 
 #include "perf_pt_private.h"
 
@@ -127,8 +128,28 @@ struct tracer_thread_args {
                         *err;               // Errors generated inside the thread.
 };
 
+// A data buffer sample indicating that new data is available in the AUX
+// buffer. This struct is not defined in a perf header, so we have to define it
+// ourselves.
+struct perf_record_aux_sample {
+    struct perf_event_header header;
+    __u64    aux_offset;
+    __u64    aux_size;
+    __u64    flags;
+    // ...
+    // More variable-sized data follows, but we don't use it.
+};
+
+// The format of the data returned by read(2) on a Perf file descriptor.
+// Note that the size of this will change if you change the Perf `read_format`
+// config field (more fields become available).
+struct read_format {
+    __u64 value;
+};
 
 // Private prototypes.
+static bool handle_sample(void *, struct perf_event_mmap_page *, struct
+                          perf_pt_trace *, void *, struct perf_pt_cerror *);
 static bool read_aux(void *, struct perf_event_mmap_page *,
                      struct perf_pt_trace *, struct perf_pt_cerror *);
 static bool poll_loop(int, int, struct perf_event_mmap_page *, void *,
@@ -141,6 +162,92 @@ struct tracer_ctx *perf_pt_init_tracer(struct tracer_conf *, struct perf_pt_cerr
 bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct perf_pt_cerror *);
 bool perf_pt_stop_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
 bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
+
+
+/*
+ * Called when the poll(2) loop is woken up with a POLL_IN. Samples are read
+ * from the Perf data buffer and an action is invoked for each depending its
+ * type.
+ *
+ * Returns true on success, or false otherwise.
+ */
+static bool
+handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
+              struct perf_pt_trace *trace, void *data_tmp,
+              struct perf_pt_cerror *err)
+{
+    // We need to use atomics with orderings to protect against 2 cases.
+    //
+    // 1) It must not be possible to read the data buffer before the most
+    //    recent head is obtained. This would mean that we may read nothing when
+    //    there is really data available.
+    //
+    // 2) We must ensure that we have already copied out of the data buffer
+    //    before we update the tail. Failure to do so would allow the kernel to
+    //    re-use the space we have just "marked free" before we copied it.
+    //
+    // The initial load of the tail is relaxed since we are the only thread
+    // mutating it and we don't mind variations on the ordering.
+    //
+    // See the following comment in the Linux kernel sources for more:
+    // https://github.com/torvalds/linux/blob/3be4aaf4e2d3eb95cce7835e8df797ae65ae5ac1/kernel/events/ring_buffer.c#L60-L85
+    void *data = (void *) hdr + hdr->data_offset;
+    __u64 head_monotonic =
+            atomic_load_explicit((_Atomic __u64 *) &hdr->data_head,
+                                 memory_order_acquire);
+    __u64 size = hdr->data_size; // No atomic load. Constant value.
+    __u64 head = head_monotonic % size; // Head must be manually wrapped.
+    __u64 tail = atomic_load_explicit((_Atomic __u64 *) &hdr->data_tail,
+                                      memory_order_relaxed);
+
+    // Copy samples out, removing wrap in the process.
+    void *data_tmp_end = data_tmp;
+    if (tail <= head) {
+        // Not wrapped.
+        memcpy(data_tmp, data + tail, head - tail);
+        data_tmp_end += head - tail;
+    } else {
+        // Wrapped.
+        memcpy(data_tmp, data + tail, size - tail);
+        data_tmp_end += size - tail;
+        memcpy(data_tmp + size - tail, data, head);
+        data_tmp_end += head;
+    }
+    atomic_store_explicit((_Atomic __u64 *) &hdr->data_tail, head, memory_order_relaxed);
+
+    void *next_sample = data_tmp;
+    while (next_sample != data_tmp_end) {
+        struct perf_event_header *sample_hdr = next_sample;
+        struct perf_record_aux_sample *rec_aux_sample;
+        switch (sample_hdr->type) {
+        case PERF_RECORD_AUX:
+                // Data was written to the AUX buffer.
+                rec_aux_sample = next_sample;
+                // Check that the data written into the AUX buffer was not
+                // truncated. If it was, then we didn't read out of the data buffer
+                // quickly/frequently enough.
+                if (rec_aux_sample->flags & PERF_AUX_FLAG_TRUNCATED) {
+                    perf_pt_set_err(err, perf_pt_cerror_ipt, pte_overflow);
+                    return false;
+                }
+                if (read_aux(aux_buf, hdr, trace, err) == false) {
+                    return false;
+                }
+                break;
+            case PERF_RECORD_LOST:
+                perf_pt_set_err(err, perf_pt_cerror_ipt, pte_overflow);
+                return false;
+                break;
+            case PERF_RECORD_LOST_SAMPLES:
+                // Shouldn't happen with PT.
+                errx(EXIT_FAILURE, "Unexpected PERF_RECORD_LOST_SAMPLES sample");
+                break;
+        }
+        next_sample += sample_hdr->size;
+    }
+
+    return true;
+}
 
 /*
  * Read data out of the AUX buffer.
@@ -221,6 +328,14 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
         {stop_fd,   POLLHUP,            0}
     };
 
+    // Temporary space for new samples in the data buffer.
+    void *data_tmp = malloc(mmap_hdr->data_size);
+    if (data_tmp == NULL) {
+        perf_pt_set_err(err, perf_pt_cerror_errno, errno);
+        ret = false;
+        goto done;
+    }
+
     while (1) {
         n_events = poll(pfds, 2, INFTIM);
         if (n_events == -1) {
@@ -232,7 +347,22 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
         // POLLIN on pfds[0]: Overflow event on either the Perf AUX or data buffer.
         // POLLHUP on pfds[1]: Tracer stopped by parent.
         if ((pfds[0].revents & POLLIN) || (pfds[1].revents & POLLHUP)) {
-            read_aux(aux, mmap_hdr, trace, err);
+            // Read from the Perf file descriptor.
+            // We don't actually use any of what we read, but it's probably
+            // best that we drain the fd anyway.
+            struct read_format fd_data;
+            if (pfds[0].revents & POLLIN) {
+                if (read(perf_fd, &fd_data, sizeof(fd_data)) == -1) {
+                    perf_pt_set_err(err, perf_pt_cerror_errno, errno);
+                    ret = false;
+                    break;
+                }
+            }
+
+            if (!handle_sample(aux, mmap_hdr, trace, data_tmp, err)) {
+                ret = false;
+                break;
+            }
 
             if (pfds[1].revents & POLLHUP) {
                 break;
@@ -246,6 +376,10 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
     }
 
 done:
+    if (data_tmp != NULL) {
+        free(data_tmp);
+    }
+
     return ret;
 }
 
@@ -290,8 +424,11 @@ open_perf(pid_t target_tid, size_t aux_bufsize, struct perf_pt_cerror *err) {
     // No skid.
     attr.precise_ip = 3;
 
-    attr.watermark = 0;
-    attr.wakeup_events = 1;
+    // Notify for every sample.
+    attr.watermark = 1;
+    attr.wakeup_watermark = 1;
+
+    // Generate a PERF_RECORD_AUX sample when the AUX buffer is almost full.
     attr.aux_watermark = (size_t) ((double) aux_bufsize * getpagesize()) * AUX_BUF_WAKE_RATIO;
 
     // Acquire file descriptor through which to talk to Intel PT. This syscall
--- a/src/backends/perf_pt/decode.c
+++ b/src/backends/perf_pt/decode.c
@@ -191,7 +191,7 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
         *addr = 0;
         return true;
     }
-    assert(*decoder_status == 0);
+    assert((*decoder_status == 0) || (*decoder_status == pts_ip_suppressed));
 
     // Now fetch the block information.
     struct pt_block block;
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -797,6 +797,14 @@ mod tests {
         }
     }
 
+    // Check that a shorter trace yields fewer blocks.
+    #[test]
+    fn test_block_iterator4() {
+        let tracer1 = default_tracer();
+        let tracer2 = default_tracer();
+        test_helpers::test_ten_times_as_many_blocks(tracer1, tracer2);
+    }
+
     // Check that a long trace causes the trace buffer to reallocate.
     #[test]
     fn test_relloc_trace_buf1() {
@@ -813,14 +821,6 @@ mod tests {
         assert!(trace.capacity() > start_bufsize);
         println!("CAP: {}", trace.capacity());
         tracer.destroy().unwrap();
-    }
-
-    // Check that a shorter trace yields fewer blocks.
-    #[test]
-    fn test_block_iterator4() {
-        let tracer1 = default_tracer();
-        let tracer2 = default_tracer();
-        test_helpers::test_ten_times_as_many_blocks(tracer1, tracer2);
     }
 
     // Check that a block iterator returns none after an error.


src/backends/perf_pt/collect.c
```diff
@@ -127,8 +128,28 @@ struct tracer_thread_args {
...
+struct perf_record_aux_sample {
+    struct perf_event_header header;
+    __u64    aux_offset;
+    __u64    aux_size;
+    __u64    flags;
+    // ...
+    // More variable-sized data follows, but we don't use it.
+};
+
+// The data structure which is read(2) out of the Perf file descriptor.
```
ltratt: This is a weird sentence -- "reading out of a file descriptor" sounds odd. Also, is this the *only* thing we ever read from perf? Does everything else just get dumped into a memory buffer?

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
 bool perf_pt_stop_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
 bool perf_pt_free_tracer(struct tracer_ctx *tr_ctx, struct perf_pt_cerror *);
 
+
+/*
+ * Reads samples from the data buffer and acts upon them appropriately.
```
ltratt: This sentence almost doesn't say anything. Can we be a bit more specific?

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+ *
+ * Returns true on success, or false otherwise.
+ */
+static bool
+handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
+         struct perf_pt_trace *trace, void *data_tmp,
```
ltratt: Not sure I understand the indentation here.

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+         struct perf_pt_cerror *err)
+{
+    bool ret = true;
```
ltratt: Given that the `clean:` target just does `return ret`, I think this variable, and the goto label, are obfuscatory.

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+
+    // We use atomic loads and stores with orderings to avoid concurrency
+    // issues WRT data buffer state mutably shared between us and the kernel.
+    // Without these, for example the compiler might be able to load data
+    // from `data_buf` before reading from the `hdr->data_head`.
+    void *data = (void *) hdr + hdr->data_offset;
+    __u64 head_monotonic =
+            atomic_load_explicit((_Atomic __u64 *) &hdr->data_head,
+                                 memory_order_acquire);
```
ltratt: Given that this value doesn't appear to have a dependency on anything else, shouldn't this be a relaxed load? Or is there some sort of implicit ordering with something else?

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+                // Check that the data written into the AUX buffer was not
+                // truncated. If it was, then we didn't read out of the data buffer
+                // quickly/frequently enough.
+                if (rec_aux_sample->flags & PERF_AUX_FLAG_TRUNCATED) {
+                    perf_pt_set_err(err, perf_pt_cerror_ipt, pte_overflow);
+                    ret = false;
+                    goto clean;
+                }
+                if (read_aux(aux_buf, hdr, trace, err) == false) {
+                    return false;
```
ltratt: Not even using the `goto` label ;)

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,84 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+                }
+                break;
+            case PERF_RECORD_LOST:
+                perf_pt_set_err(err, perf_pt_cerror_ipt, pte_overflow);
+                ret = false;
+                goto clean;
+                break;
+            case PERF_RECORD_LOST_SAMPLES:
+                // Shouldn't happen with PT.
+                assert(false);
```
ltratt: `assert` could be removed by a compiler, so this isn't a safe way of guarding against an exceptional event AFAIK.

src/backends/perf_pt/collect.c
```diff
@@ -221,6 +320,14 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
         {stop_fd,   POLLHUP,            0}
     };
 
+    // Temproary copy space for new samples in the data buffer.
```
ltratt: Typo: "temporary". I've also never heard the term "copy space" before!

src/backends/perf_pt/collect.c
```diff
@@ -232,7 +339,22 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
         // POLLIN on pfds[0]: Overflow event on either the Perf AUX or data buffer.
         // POLLHUP on pfds[1]: Tracer stopped by parent.
         if ((pfds[0].revents & POLLIN) || (pfds[1].revents & POLLHUP)) {
-            read_aux(aux, mmap_hdr, trace, err);
+            // Read from the Perf file descriptor.
+            // We don't actually use any of what we read, but it's probably
```
ltratt: Probably?

src/backends/perf_pt/collect.c
```diff
@@ -246,6 +368,10 @@ poll_loop(int perf_fd, int stop_fd, struct perf_event_mmap_page *mmap_hdr,
     }
 
 done:
+    if (data_tmp != NULL) {
```
ltratt: I think `data_type` is always `!= NULL` at this point.

src/backends/perf_pt/collect.c
```diff
@@ -292,14 +292,16 @@ read_aux(void *aux_buf, struct perf_event_mmap_page *hdr,
         trace->buf = new_buf;
     }
 
-    // Finally copy the AUX buffer into the trace buffer.
+    // Finally append a copy the new AUX data onto the trace storage buffer.
```
ltratt: Slightly awkward phrasing with "onto".

src/backends/perf_pt/collect.c
vext01: It's the only thing ever read using the `read` system call, yes. But as you have said, there are two `mmap` regions which are being filled. What we read out of the fd tells us how/when to look at these buffers. It's complicated.

I'm not sure what's wrong with "reading out of a file descriptor" though. Do you prefer "reading from a file descriptor"?

For comparison, here's how the `perf_event_open` (dryly) uses prose for this:
> This field specifies the format of the data returned by read(2) on a perf_event_open() file descriptor.

src/backends/perf_pt/collect.c
vext01: It's a bit like the interpreter loop in a VM. It tells us what actions to perform. Did you have something specific in mind?

src/backends/perf_pt/collect.c
ltratt: Maybe we could say something like "The data-structure returned via the perf file descriptor, which then allows us to make sense of the memory buffers"?

src/backends/perf_pt/collect.c
vext01: Not sure I do either! Sorry I've been editing a mix of code styles recently since I've been working inside libipt!

src/backends/perf_pt/collect.c
ltratt: I didn't really understand what the function does so... no ;)

src/backends/perf_pt/collect.c
vext01: I don't like "returned" here. I can think of a couple of ways to get data from an fd, and not just via  `read(2)`. E.g. `ioctl(2)`.

src/backends/perf_pt/collect.c
vext01: OK, I'll try to find something more explicit.

src/backends/perf_pt/collect.c
ltratt: I don't think you can read data via `ioctl`?

Anyway, maybe "The data-structure read via the perf file ..."?

src/backends/perf_pt/collect.c
vext01: I think you are right. An earlier version probably had cleanup, which is why this has happened.

src/backends/perf_pt/collect.c
vext01: As I understand it, we don't want the load of the buffer to come before the load of the head. It's the same logic as in `read_aux`.

src/backends/perf_pt/collect.c
vext01: Oops. Well, we are killing the label, so this is actually OK. But we shouldn't mix for sure.

src/backends/perf_pt/collect.c
vext01: Hrm, I didn't know that:

>      The assert() macro may be removed at compile time with the cc(1) option -DNDEBUG.

So we should review all uses of assert then.

src/backends/perf_pt/collect.c
vext01: But you like it ;)

ltratt: Um...

src/backends/perf_pt/collect.c
vext01: It's a gut feeling. No one says I should empty the fd, but something tells me we can't buffer it up forever. Don't you agree?

src/backends/perf_pt/collect.c
vext01: If allocation fails, `data_tmp` is `NULL`.

src/backends/perf_pt/collect.c
vext01: "append a copy the new AUX data to the trace storage buffer."?

src/backends/perf_pt/collect.c
vext01: It depends on the ioctl in question, but you certainly can get stuff out:
```
An ioctl request has encoded in it whether the argument is an “in” parameter or “out” parameter
```

> "The data-structure read via the perf file ..."?

WFM

src/backends/perf_pt/collect.c
ltratt: I don't remember `read_aux` :) Still, this is another chance to think about this again.

Right, so there's a dependency between `hdr->data_head` and `data` right? I presume that only PT changes `hdr->data_head` but only we change `hdr->data_tail`?

src/backends/perf_pt/collect.c
ltratt: I guess I also partly wondered why we're throwing everything away...

src/backends/perf_pt/collect.c
ltratt: How about just "Add the AUX data to the end of the trace buffer" or similar?

src/backends/perf_pt/collect.c
vext01: correct.

src/backends/perf_pt/collect.c
vext01: OK. In a later PR I think I'm going to add a `unexpected()` function/macro which prints a message to stderr before crashing out with `errx`. We can look at all existing asserts and decide if we want to replace them.

vext01: Actually errx already prints, so probably no need.

src/backends/perf_pt/collect.c
ltratt: OK, can we note this down in a comment? Because now I understand why you have acquire/release loads.

vext01: I'm not sure it's necessary as it's well documented in the `perf_event_open` manual. If you feel strongly, OK.

ltratt: Given that it's over 2000 lines long, that man page may also contain the secret to eternal life...

As you know, in general, I'm not a fan of too many comments. But anything to do with atomics has to be treated with the utmost caution. The problem is that the programmer's intent is not generally obvious from the code (I had to guess the above). And then I'd need to know what to look for in the man page to see the comment `On  SMP-capable  platforms,  after  reading the data_head value, space should issue an rmb().`.

vext01: Ah, well that part is already mentioned in a comment. I thought you meant you wanted to add:
> only PT changes hdr->data_head but only we change hdr->data_tail

vext01: Currently we have:
```
      // We use atomic loads and stores with orderings to avoid concurrency          
      // issues WRT data buffer state mutably shared between us and the kernel.   
      // Without these, for example the compiler might be able to load data        
      // from `data_buf` before reading from the `hdr->data_head`.             
```

Does that cover it for you? If not, how should we change it to be clearer?

ltratt: It's not really sufficient. For example, it doesn't say what the dependencies we have to observe actually are: it has just a very vague line about what the compiler could do (who cares what it could do? I only care about the *bad* things it could do!).

I would prefer to be clear here with something like: "There is a dependency between hdr->data_head (which is only written to by PT) and data (whose contents are also written to by PT): we need to use an acquire barrier to ensure that the former is a valid pointer into the latter. Although hdr->data_tail is only written to by us, since we may have more than one of our own threads writing to it, we also need to use an atomic to ensure that is synchronised between threads."

BTW, writing that out makes me realise that we probably only need "relax" ordering on `hdr->tail`...

vext01: > since we may have more than one of our own threads writing to it

That would cause chaos. I don't think we can trace the same tid twice.

> we need to use an acquire barrier to ensure that the former is a valid pointer

The pointer will be valid, but maybe not up to date.

I'm not sure we are agreeing. As I see it there's a dependency between data and *both* head and tail. We want to make sure that we always have up to date values of those before looking into the data buffer at those offsets.

vext01: > BTW, writing that out makes me realise that we probably only need "relax" ordering on hdr->tail

I don't understand, can you explain?

vext01: Because we are the only thread to write the tail, we can rely on program order to get it right?

ltratt: >That would cause chaos. I don't think we can trace the same tid twice.

I'm not saying that we trace the same TID twice: I'm saying that, I think, thread t1 and thread t2 might read from the PT buffer at different times.

>As I see it there's a dependency between data and both head and tail.

AFAICS *we* never write to `data` or `hdr->data_head`, and we are fully in control of `hdr->data_tail`. If so, there's no write dependency between the first two and the third.

vext01: > I'm not saying that we trace the same TID twice: I'm saying that, I think, thread t1 and thread t2 might read from the PT buffer at different times.

Don't worry about this. If we are tracing two different TIDs then t1 and t2 would each have their own mmaped buffers. Otherwise the packets would interleave and you wouldn't know which packet is for which thread.

> AFAICS we never write to data or hdr->data_head, and we are fully in control of hdr->data_tail. If so, there's no write dependency between the first two and the third.

I'm confused.

vext01: Maybe I understand this.

vext01: Wait, the kernel still has to read the tail so as to know not to overwrite data past this point. The release causes the other cores to see the new value at the time they acquire the tail.

ltratt: Then you still only need relaxed ordering, since `tail` stands on its own (we haven't written to `head` or `data`).

vext01: I'm totally confused.

I suggest we keep it as is, as it's more conservative. You can try and explain it to me at the next thrash ;)

ltratt: My suggestion is the following:

```
let head = atomic_read(&hdr->data_head, ACQUIRE);
let tail = atomic_read(&hdr->data_tail, RELAXED);
...
<do the memcpy stuff>
...
atomic_write(&hdr->data_tail, head, RELAXED);
```

We have to use ACQUIRE for `hdr->data_head` because we want to make sure that everything in `data` is upto date (and the kernel's write to `hdr->data_head` serves a sort-of lock to guarantee that). However, we can get away with RELAXED for `hdr->data_tail` because we're the only people to write to it (and it doesn't need a dependency on `data`).

Feel free to challenge me on this!

vext01: What I don't get is that the atomic read will get the latest version of the head pointer regardless of the ordering. AFAICS the ordering just changes where the read could happen. We know that the atomic read must happen before `if (tail <= head) {` because it uses `head`. So why would it matter if we load the tail or the buffer size before the head? They are the only remaining memory operations left which could be reordered. Do you follow?

vext01: (All uses of `data` come after `if (tail <= head)`)

ltratt: The point is we want everything we read from `data` to be up to date. If we read `head` before everything to `data` is flushed, we might be in trouble.

vext01: Sorry if I'm being dumb, but why does reading head with acquire order guarantee the data buffer to be flushed?

vext01: If we assume that the kernel uses the right memory ordering to ensure that the head pointer is only updated after the data is written, then any atomic read of the head which occurs prior to the memcpy should give the correct result.

ltratt: Both sides have to use the right ordering (and, I admit, I'm being pessimistic: I don't know *when* the kernel updates `head`, so I'm assuming the worst case). If they used release but we used relaxed, we would read `tail` correctly but maybe not `data`.

src/backends/perf_pt/collect.c
vext01: You only read counters out of the fd, which we don't need for our use case.

src/backends/perf_pt/collect.c
vext01: I think it's safe to assume the kernel does it right.

vext01: Looking at the perf source code, this is how they read the head and write the tail:
https://github.com/torvalds/linux/blob/3be4aaf4e2d3eb95cce7835e8df797ae65ae5ac1/tools/perf/util/mmap.h#L71-L88

So here they are using not atomics, but barriers.

The meanings of those barriers are:

>mb() | A full  system memory barrier. All memory operations before the mb() in the  instruction stream will be committed before any operations after the  mb() are committed. This ordering will be visible to all bus masters in  the system. It will also ensure the order in which accesses from a  single processor reaches slave devices.

> rmb() | Like  mb(), but only guarantees ordering between read accesses. That is, all  read operations before an rmb() will be committed before any read  operations after the rmb().

I don't know that these barriers map well to the ordered atomic ops that we want to use though.

vext01: It seems we are the only one on the entire of github using C11 atomics on the perf buffers :\

vext01: I think I can offer reasoning as to why the code in-tree is safe.

There are two things we need to prevent:
 * Reading the data buffer before the most recent head is obtained. This would mean that we may read nothing when there is really data available.
 * Updating the tail before we have read data out of the data buffer. This would mean the kernel is free to overwrite the data we have not read yet.

Then:
 * By reading the head first in acquire order, it means we simply cannot read the data prior to holding the latest head value.
 * By writing the tail with release ordering, it means the memcpys cannot be moved after the write to the tail.

ltratt: The thing is, we don't know when head and data are written with respect to each other. What happens if head is written first and then data is updated? [I think it's unlikely that this is the way things are done, but no document seems to say it couldn't happen, so we have to be conservative.]

vext01: The above says nothing about what ordering should be used for the one remaining atomic, the load of the tail.

I think since we are the only writer of the tail, the ordering doesn't matter. The read just has to happen before the write, which is a given for the same atomic variable, and thus can be relaxed.

(elided)

```
    void *data = (void *) hdr + hdr->data_offset;
    __u64 head_monotonic =
            atomic_load_explicit((_Atomic __u64 *) &hdr->data_head,
                                 memory_order_acquire);
    __u64 size = hdr->data_size; // No atomic load. Not mutated in-kernel.
    __u64 head = head_monotonic % size; // Head must be manually wrapped.
    __u64 tail = atomic_load_explicit((_Atomic __u64 *) &hdr->data_tail,
                                 memory_order_relaxed); // <<<<< CHANGED ORDER

    // Copy samples out, removing wrap in the process.
    void *data_tmp_end = data_tmp;
    if (tail <= head) {
        // Not wrapped.
        memcpy(data_tmp, data + tail, head - tail);
        data_tmp_end += head - tail;
    } else {
        // Wrapped.
        memcpy(data_tmp, data + tail, size - tail);
        data_tmp_end += size - tail;
        memcpy(data_tmp + size - tail, data, head);
        data_tmp_end += head;
    }
atomic_store_explicit((_Atomic __u64 *) &hdr->data_tail, head, memory_order_release);
```

Do you agree?

vext01: > What happens if head is written first and then data is updated?

That's called a kernel bug. And if you really are worried I can find the code responsible in the kernel.

vext01: Both sides have to cooperate or neither side can win.

ltratt: >That's called a kernel bug. And if you really are worried I can find the code responsible in the kernel.

Why is this a bug?

ltratt: Of course it could work that way. It would be odd, but unless the spec disallows it, one can't assume it will never happen. Indeed, unless the spec says there's an ordering between head and data, we can't assume anything. Hence why a full barrier is the only safe option. [I'm not saying I like this outcome, but I think it's the one that's least likely to bite us in the rear end in the future.]

vext01: > Of course it could work that way.

I don't follow. How would you know if it was safe to read the data behind a mutex if the value of a mutex had no bearing on if the data had been written?

ltratt: There is no mutex here though.

vext01: Its the same. How would you know how much data to read from the buffer if the kernel didn't guarantee that all data up to `head` was ready? That would be a logic bug and there's nothing the user thread could ever do to be correct under that scenario.

vext01: I've pushed a change with a big comment explaining my reasoning. Does it work for you?

vext01:
Hopefully that's all comments addressed.

src/backends/perf_pt/collect.c
ltratt: Sort-of, but only because the man page says:
```
       data_head
              This points to the head of the data section.  The value continu‐
              ously  increases, it does not wrap.  The value needs to be manu‐
              ally wrapped by the size of the mmap buffer before accessing the
              samples.

              On  SMP-capable  platforms,  after  reading the data_head value,
              user space should issue an rmb().
```
And because it doesn't say anything like that for data_tail. In other words, the *only* reason this is correct is because the spec says it is. I think I would prefer to point the reader at the spec rather than engage in reasoning about the system's internals.

[FWIW, I think the docs for this are terrible. They could have saved us much heartache if they'd documented stuff correctly. Oh well.]

vext01: What spec are we talking about here? The C11 language spec?

I'm actually very happy with the comment I've written as I feel it's the first time I've understood this stuff properly. I feel that in a 6 months time I'd be able to understand based on this comment.

vext01:
Trimmed down the comment and added a link. Works for you?

vext01:
Maybe we are good to squash?

src/backends/perf_pt/collect.c
```diff
@@ -187,7 +187,7 @@ handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
     //    re-use the space we have just "marked free" before we copied it.
     //
```
ltratt: Now you've got two blank lines?

vext01: Eh? The comment in full is:
```
      // We need to use atmoics with orderings to protect against 2 cases.    
      //    
      // 1) It must not be possible to read the data buffer before the most    
      //    recent head is obtained. This would mean that we may read nothing when
      //    there is really data available.    
      //    
      // 2) We must ensure that we have already copied out of the data buffer    
      //    before we update the tail. Failure to do so would allow the kernel to 
      //    re-use the space we have just "marked free" before we copied it.    
      //    
      // The initial load of the tail is relaxed since we are the only thread        
      // mutating it and we don't mind variations on the ordering.    
      //    
      // See the following comment in the Linux kernel sources for more:    
      // https://github.com/torvalds/linux/blob/3be4aaf4e2d3eb95cce7835e8df797ae65ae5ac1/kernel/events/ring_buffer.c#L60-L85
```

Am I being silly?

ltratt: Nope, I seemed to have misread the diff. Sorry.

ltratt:
OK, let's squash.

vext01:
Here we go. I put the atomics change for read_aux in its own commit for clarity.

ltratt:
I think that would be best squashed away: it doesn't seem helpful to have an incorrect commit in the same PR as the commit that fixes it.

vext01:
Like this?

src/backends/perf_pt/collect.c
```diff
@@ -142,6 +163,92 @@ bool perf_pt_start_tracer(struct tracer_ctx *, struct perf_pt_trace *, struct pe
...
+ * type.
+ *
+ * Returns true on success, or false otherwise.
+ */
+static bool
+handle_sample(void *aux_buf, struct perf_event_mmap_page *hdr,
+              struct perf_pt_trace *trace, void *data_tmp,
+              struct perf_pt_cerror *err)
+{
+    // We need to use atmoics with orderings to protect against 2 cases.
```
ltratt: Typo: atomics.

vext01: argh!

ltratt:
Yes. Just spotted a minor typo. Just fix that in a force push and then I'll merge.

vext01:
fixed and forced.

PR #62, opened 2018-04-26
Allow interrupted traces

vext01:
Now that we understand [this](https://github.com/01org/processor-trace/issues/43), this branch lifts the restriction that test traces must be uninterrupted by asynchronous events such as context switches. This means we can test larger traces.

There's a couple of other bits in here too. See individual commits.

OK?


vext01:
Oh, not that it would change anything with the current incarnation of travis, but I suppose CI should run the ignored tests too?

src/backends/perf_pt/mod.rs
```diff
@@ -785,6 +785,15 @@ mod tests {
         test_helpers::test_ten_times_as_many_blocks(tracer1, tracer2);
     }
 
+    // Check that our block decoder agrees ptxed on long trace.
+    // XXX We use an even higher iteration count once our decoder uses a libipt image cache.
+    #[ignore] // Decoding long traces is slow.
```
ltratt: How slow is slow?

vext01: This test takes about a minute. It's long enough to annoy me during development.

ltratt:
I think we can probably squash this a bit before merging? e.g. the commit with the test in can probably be squashed into its predecessor?

PR #63, opened 2018-04-27
Review asserts

### LINKED ISSUE #61: Audit use of assert in tree.
Consider replacing with errx, possibly using a function/macro.

E.g. instead of:
```
if (something.state() == BAD) {
    errx(EXIT_STATUS, "something bad");
}
```

this:
```
xassert(something.state() == BAD, "something bad");
```

vext01:
Hi Laurie,

What do you think of this to fix #61?

You are going to ask: why a macro?
 * There's no `errx` that accepts a `va_list` (but we could have used `vfprintf`).
 * By using a macro we can expand `__FILE__` and `__LINE__` for the call-site thus "throwing me a bone" as to where to look.

Example failure output:
```
xassert failure: src/backends/perf_pt/decode.c:196
simple_example: Unexpected decoder status: 456
```

(Also moved my DEBUG macro to a more appropriate home).

Cheers


vext01:
> An assert should really be an internal consistency check (i.e. making sure that one part of our program hasn't done something that messes up another part)

I agree that that's one good use-case for an assertion. But in my opinion they are more general than that. It's a way of saying "if this isn't true, stop" (it's in the name: we *assert* what needs to be true for the program to continue). And we can use this mechanism to stop the program if (e.g.) we reach an unimplemented case.

> Those, in my opinion, should be if (err) errx("blah")

That's almost exactly what the macro does thought, isn't it?

ltratt:
The problem with the macro is that these *look* like internal consistency checks, but they're actually checks of external consistency.

vext01:
Why does it matter though? The user sees the same (a crash), and the developer (me) knows what is what. Cases which are not a consistency check clearly mention that something is "unimplemented" in the crash string.

ltratt:
It matters because other readers of the code won't have the same assumptions as you will. They're used to `assert` for internal consistency checks and `if (err) errx("blah")` for external consistency checks. Since it's much harder to read code than write it, we want to optimise for the later-reader-who-isn't-you whenever possible.

vext01:
> They're used to assert for internal consistency checks and if (err) errx("blah") for external consistency checks.

I don't understand what you mean by "external consistency check" here.

ltratt:
You're right, "external consistency check" doesn't make sense. I should have said something like "check that values obtained from the outside world are correct."

vext01:
> check that values obtained from the outside world are correct.

Yeah, I agree that `err*` is the right tool for that particular job, but is that what we are doing here?

Give me a moment...

ltratt:
Basically I'm just asking for the macro to be expanded ;)

src/backends/perf_pt/collect.c
```diff
@@ -493,7 +493,7 @@ tracer_thread(void *arg)
 
 clean:
     if (!sem_posted) {
-        assert(!ret);
+        xassert(!ret, "Expected failure return code");
```
vext01: That's checking that we only unlock the semaphore if we didn't fail. So that's an internal consistency check.

src/backends/perf_pt/decode.c
```diff
@@ -191,7 +191,8 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
         *addr = 0;
         return true;
     }
-    assert((*decoder_status == 0) || (*decoder_status == pts_ip_suppressed));
+    xassert(*decoder_status == 0 || *decoder_status == pts_ip_suppressed,
```
vext01: That's checking that we don't get any PT events here, because the earlier call to `handle_events` should have drained them. So that's an internal consistency check.

src/backends/perf_pt/decode.c
```diff
@@ -212,14 +213,15 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
     }
     // It's possible at this point that we get notified of an event in the
     // stream. This will be handled in the next call to `perf_pt_next_block`.
-    assert((*decoder_status == 0) || (*decoder_status == pts_event_pending));
+    xassert(*decoder_status == 0 || *decoder_status == pts_event_pending,
```
vext01: Same as above here.

src/backends/perf_pt/decode.c
```diff
@@ -212,14 +213,15 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
+            "Unexpected decoder status: %d", *decoder_status);
 
     // XXX A truncated block occurs when a block straddles a section boundary.
     // In this case we may need some extra logic, but this should be rare.
-    assert(block.truncated == 0);
+    xassert(block.truncated == 0, "Truncated blocks are not implemented");
```
vext01: This is an unimplemented case. Which is neither an internal consistency check, nor checking values from the user.

src/backends/perf_pt/collect.c
ltratt: Fair point: "internal" and "external" aren't good words. Yes, this is internal to our address space; but it's an external check in the sense that it's outside our source code. Also, it's expected that sometimes it might fail, and that we might (one day) be able to recover from it.

src/backends/perf_pt/decode.c
```diff
@@ -212,14 +213,15 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
 
     // A block should have at least one instruction.
-    assert(block.ninsn > 0);
+    xassert(block.ninsn > 0, "Detected a block with 0 instructions");
```
vext01: That apparently *can* happen, but I want to know when ;) So I guess that's challenging my assumption that it can't, so we can at a stretch call that an internal consistency check.

src/backends/perf_pt/decode.c
```diff
@@ -311,8 +313,7 @@ handle_events(struct pt_block_decoder *decoder, int *decoder_status, struct perf
             // didn't ask them to be emitted, e.g. TSC, STOP and CYC packets.
             // We print what packet crashed us before dying to aid debugging.
             default:
-                fprintf(stderr, "Unhandled packet event type %d\n", event.type);
-                assert(false);
+                xassert(false, "Unhandled packet event type %d", event.type);
```
vext01: That's an "unimplemented" scenario.

vext01:
OK, I've annotated all of the `xassert` cases. There are two classes of things:

 * unimplemented.
 * checking assumptions.

We could define anther macro (modularly) which separates the two?

ltratt:
I'm really not a fan of a macro for either of these cases. Is it really so bad to write these as normal checks, like other programs do?

src/backends/perf_pt/collect.c
vext01: Yup, and that's fine. When that day comes `xassert` will give me a good error message and file/line and we can make the change.

I think you are being too philosophical :)

src/backends/perf_pt/decode.c
vext01: I realise now, you didn't mean checking user values. You mean checking the return codes of external libs.

vext01:
Well, we could, but it's more tedious to read and write.

```
xassert(block.ninsn > 0, "Detected a block with 0 instructions");
```

Becomes:
```
if (block.ninsn > 0) {
    errx(EXIT_FAILURE, "error: %s:%d\n", __FILE__, __LINE__);
}
```

>  like other programs do?

Hah, you've not done enough porting. I've seen all kinds of different strategies. There is no unified view as far as I can tell. But I hope what I've proposed is on the more sensible side :)

I know you don't like macros, but IMHO this is a good use case for one.

ltratt:
Modulo that one comment, this looks fien to me.

vext01:
Oh yeah. I killed one assertion which I thought was a bit excessive. What do you think?

ltratt:
Can the code recover if that path is taken? If so, the assert is definitely best removed.

vext01:
It was a "this can't happen" assertion, so whether it could recover or not kind of irrelevant.

ltratt:
If it's a "definitely can't happen", then why do we have that case in there? If it's a "we don't think this can happen but maybe we're wrong", then are we sure the error path is correct?

vext01:
The cleanup can happen, but the assertion could not.

ltratt:
OK.

Right, time to squash!

vext01:
Let me expand upon that, just to check we agree:
```
    int sem_posted = 0;
    bool ret = true;

    ...

    if (sem_post(thr_args->tracer_init_sem) != 0) {
        perf_pt_set_err(err, perf_pt_cerror_errno, errno);
        ret = false;
        goto clean;
    }
    sem_posted = 1;

    ...

clean:
    if (!sem_posted) {
        assert(!ret); <--- I KILLED THIS
        sem_post(thr_args->tracer_init_sem);
}
```

See, `sem_posted` is only ever 0 if we failed earlier. It's kinda overkill.

Why not a bool? permission to update this too?

ltratt:
I'm not sure what the "why not a bool" is referring to. But feel free to add a commit and I'll review it (although motivation in the commit will help me :)).

vext01:
Like this. I'm 99.999% you will approve :)

vext01:
So shall I squash?

ltratt:
Yes, I like it :) Please squash!

vext01:
splat.

ltratt:
Should that second commit message be "FOR debug"?

vext01:
Ugh! Yes!

vext01:
Fixed

PR #65, opened 2018-04-27
Alllow the user to configure when to empty the AUX ringbuffer.

vext01:
I think it would be a good idea to expose this to the user as a configuration variable.

See individual commits.

Thanks

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -321,6 +321,8 @@ pub struct PerfPTConf {
     data_bufsize: size_t,
     // AUX buffer size, in pages. Must be a power of 2.
     aux_bufsize: size_t,
+    // Fraction of the AUX buffer to fill before waking up to copy data out.
+    aux_copyout_threshold: f32,
     // The initial trace buffer size (in bytes) for new traces.
     new_trace_bufsize: size_t,
 }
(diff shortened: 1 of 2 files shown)


src/backends/perf_pt/mod.rs
```diff
@@ -416,7 +427,9 @@ impl PerfPTTracer {
         if !power_of_2(config.aux_bufsize) {
             return Err(HWTracerError::BadConfig(String::from("aux_bufsize must be a positive power of 2")));
         }
-
+        if config.aux_copyout_threshold < 0. || config.aux_copyout_threshold > 1. {
```
ltratt: Shouldn't this error be in the `aux_copyout_threshold` function?

vext01: That's a copy-pasta, indeed.

vext01:
> I don't think we should bother setting configuration options until we know we'll need them.

I agree that I am speculating the need for this config option. I'm imagine that different parameters will be needed for different hardware (depending upon the speed the CPU execute code, and the speed of the memory subsystem, etc.).

If you still feel that this is unneccessary, I suggest we close this and leave the branch around until we have at-least tested on some different hardware.

Thoughts?

ltratt:
I must admit, I would prefer to do this on an as-needed basis.

vext01:
OK, closing for now.

PR #67, opened 2018-04-30
Merge partial blocks.

vext01:
We learned from Markus that blocks could end prematurely for a number of reasons. We (AFAICS) only every want complete blocks terminated with a control flow instruction.

This change makes our block iterator disregard blocks which were preceded by a partial block when deciding the next block start address that we present to the user.

Makes sense?

### DIFF
--- a/src/backends/perf_pt/decode.c
+++ b/src/backends/perf_pt/decode.c
@@ -62,6 +62,7 @@ struct load_self_image_args {
 static bool handle_events(struct pt_block_decoder *, int *, struct perf_pt_cerror *);
 static bool load_self_image(struct load_self_image_args *);
 static int load_self_image_cb(struct dl_phdr_info *, size_t, void *);
+static bool block_is_terminated(struct pt_block *);
 
 // Public prototypes.
 void *perf_pt_init_block_decoder(void *, uint64_t, int, char *, int *,
@@ -195,41 +196,60 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
         panic("Unexpected decoder status: %d", *decoder_status);
     }
 
-    // Now fetch the block information.
+    // The libipt block decoder may return a partial block (it could have been
+    // interrupted for example). We abstract this detail away. Using a loop we
+    // record (and eventually return) the address of the first block we see,
+    // then keep decoding more blocks until we see a properly terminated block.
     struct pt_block block;
-    *decoder_status = pt_blk_next(decoder, &block, sizeof(block));
-    // Other +ve decoder status codes can arise here. We ignore them for now,
-    // and let them be detected by handle_events() above when we are next
-    // called.
-    if (*decoder_status < 0) {
+    block.iclass = ptic_other;
+    bool first_block = true;
+    while (!block_is_terminated(&block)) {
+        if (handle_events(decoder, decoder_status, err) != true) {
+            // handle_events will have already called perf_pt_set_err().
+            return false;
+        } else if (*decoder_status & pts_eos) {
+            // End of stream.
+            *addr = 0;
+            return true;
+        }
+        // It's possible at this point that we get notified of an event in the
+        // stream. This will be handled in the next call to `perf_pt_next_block`.
+        if ((*decoder_status != 0) && (*decoder_status != pts_event_pending)) {
+            panic("Unexpected decoder status: %d", *decoder_status);
+        }
+
+        *decoder_status = pt_blk_next(decoder, &block, sizeof(block));
+        // Other +ve decoder status codes can arise here. We ignore them for now,
+        // and let them be detected by handle_events() above when we are next
+        // called.
         if (*decoder_status == -pte_eos) {
             // End of stream is flagged as an error in the case of pt_blk_next().
             *addr = 0;
             return true;
-        } else {
+        } else if (*decoder_status < 0) {
             // A real error.
             perf_pt_set_err(err, perf_pt_cerror_ipt, -*decoder_status);
             return false;
         }
-    }
-    // It's possible at this point that we get notified of an event in the
-    // stream. This will be handled in the next call to `perf_pt_next_block`.
-    if ((*decoder_status != 0) && (*decoder_status != pts_event_pending)) {
-        panic("Unexpected decoder status: %d", *decoder_status);
+
+        // XXX A truncated block occurs when a block straddles a section boundary.
+        // In this case we may need some extra logic, but this should be rare.
+        if (block.truncated != 0) {
+            panic("Truncated blocks are not implemented");
+        }
+
+        // A block should have at least one instruction.
+        if (block.ninsn == 0) {
+            panic("Detected a block with 0 instructions");
+        }
+
+        if (first_block) {
+            // The block start address we report back to the user.
+            *addr = block.ip;
+            first_block = false;
+        }
     }
 
-    // XXX A truncated block occurs when a block straddles a section boundary.
-    // In this case we may need some extra logic, but this should be rare.
-    if (block.truncated != 0) {
-        panic("Truncated blocks are not implemented");
-    }
-
-    // A block should have at least one instruction.
-    if (block.ninsn == 0) {
-        panic("Detected a block with 0 instructions");
-    }
-
-    *addr = block.ip;
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/backends/perf_pt/mod.rs
```diff
@@ -616,6 +612,22 @@ mod tests {
...
+    /*
+     * Determine if the given x86_64 assembler mnemonic should terminate a block.
+     *
+     * Mnemonic assumed to be lower case.
+     */
+    fn instr_terminates_block(instr: &str) -> bool {
+        assert!(!instr.starts_with("["));
+        match instr {
+            // JMP or Jcc are the only instructions beginning with 'j'.
+            m if m.starts_with("j") => true,
```
vext01: Let me point out here that `Jcc` is not one instruction. `cc` is a place-holder for a condition, so there's `jg`, `jge`, `jz`... Nearly a hundred variations in total. That's why I didn't enumerate them all.

vext01: Sorry, that's nearly a hundred *encodings* (depending on the types of the operands). So we'd have to list less, looks like 35 or so, if we didn't wan't to use this catch-all case.

src/backends/perf_pt/decode.c
```diff
@@ -195,41 +196,63 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
+        if ((*decoder_status != 0) && (*decoder_status != pts_event_pending)) {
+            panic("Unexpected decoder status: %d", *decoder_status);
+        }
 
-    // XXX A truncated block occurs when a block straddles a section boundary.
-    // In this case we may need some extra logic, but this should be rare.
-    if (block.truncated != 0) {
-        panic("Truncated blocks are not implemented");
-    }
+        // Now fetch the block information.
```
ltratt: Redundant comment.

src/backends/perf_pt/decode.c
```diff
@@ -195,41 +196,63 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
+        *decoder_status = pt_blk_next(decoder, &block, sizeof(block));
+        // Other +ve decoder status codes can arise here. We ignore them for now,
+        // and let them be detected by handle_events() above when we are next
+        // called.
+        if (*decoder_status < 0) {
```
ltratt: This is a really weird `if` block. Couldn't it just be:
```c
if (*decode_status == -pte_eos) { ... }
else if (*decode_status < 0) { ... }
```

src/backends/perf_pt/decode.c
```diff
@@ -195,41 +196,63 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
+            }
+        }
+
+        // XXX A truncated block occurs when a block straddles a section boundary.
+        // In this case we may need some extra logic, but this should be rare.
+        if (block.truncated != 0) {
+            panic("Truncated blocks are not implemented");
+        }
+
+        // A block should have at least one instruction.
```
ltratt: Is this something we need to check for here?

src/backends/perf_pt/mod.rs
```diff
@@ -616,6 +612,22 @@ mod tests {
         (ptxed_args, vdso_tempfile)
     }
 
+    /*
+     * Determine if the given x86_64 assembler mnemonic should terminate a block.
+     *
+     * Mnemonic assumed to be lower case.
+     */
+    fn instr_terminates_block(instr: &str) -> bool {
+        assert!(!instr.starts_with("["));
```
ltratt: Why is this assert here?

src/backends/perf_pt/mod.rs
```diff
@@ -616,6 +612,22 @@ mod tests {
         (ptxed_args, vdso_tempfile)
     }
 
+    /*
+     * Determine if the given x86_64 assembler mnemonic should terminate a block.
+     *
+     * Mnemonic assumed to be lower case.
```
ltratt: Why?

src/backends/perf_pt/decode.c
```diff
@@ -325,6 +348,36 @@ handle_events(struct pt_block_decoder *decoder, int *decoder_status, struct perf
...
 
+/*
+ * Decides if a block is terminated by a control flow dispatch.
+ *
+ * This is used to decide if libipt gave us a partial block or not.
+ */
+static bool
+block_is_terminated(struct pt_block *blk)
+{
+    bool ret;
```
ltratt: This seems a tedious way of just saying `return true` below?

src/backends/perf_pt/decode.c
vext01: It was there before. But I agree we can kill it.

src/backends/perf_pt/decode.c
vext01: This is also in master (just indented). Yes, I think it could be clearer as you suggested.

src/backends/perf_pt/decode.c
vext01: I'm not sure what you mean.

ltratt: Why are we checking this?

src/backends/perf_pt/decode.c
vext01: I started out with early returns, but you end up with unreachable code and another panic.

What do you prefer?

src/backends/perf_pt/mod.rs
vext01: Why is it assumed lower case? `ptxed` prints instructions in lower case and that's where the input comes from.

I did consider explicitly lower-casing, but this will slow the tests down, especially the long one.

Note that this is testing-only code. In the main library, we use integer comparisons. We can't use them here because we are scraping the output of a command.

src/backends/perf_pt/mod.rs
vext01: It's debug code. I'll kill it.

(this branch predates the panic change).

src/backends/perf_pt/decode.c
vext01: I think it can't happen, but I want to know if it can.

src/backends/perf_pt/decode.c
ltratt: Ah, one of those :/ OK, leave as-is.

src/backends/perf_pt/mod.rs
ltratt: Maybe just add an `assert` to check for lower case rather than documenting it?

src/backends/perf_pt/decode.c
ltratt: And if it happened... why would this be the right place to catch it?

vext01: If I moved it outside the loop, I'd only be checking blocks which are the start of a fully-terminated block, and not all blocks (partial or otherwise).

Is that what you mean? I'm not 100% sure what you are asking :P

ltratt: I understand why some parts of the code might need to check this invariant holds, but this doesn't seem like the right place for it to be checked (to me, at least).

src/backends/perf_pt/mod.rs
vext01: Are strings counted in Rust? Otherwise we are still going to slow tests down.

src/backends/perf_pt/decode.c
vext01: Roger.

src/backends/perf_pt/mod.rs
ltratt: Counted?

vext01: Wait, that's rubbish, what I've just said. It doesn't matter, the assertion will force a linear pass over each line of assembler code.

vext01: Oddly enough, adding
```
assert_eq!(instr.find(|c: char| !c.is_lowercase()), None);
```

Doesn't slow the long benchmark down. Would that work for you?

src/backends/perf_pt/decode.c
vext01: Where would you put it?

ltratt: I don't know. It feels like it's a concern of the user of this code to me, but I could be persuaded otherwise.

vext01: The user won't see every block libipt generates.

Imagine we get two blocks from libipt:
```
[block]
mov a, b
add b, c
[block]
jmp 0xdeadbeef
```

The first block is partial, so users of our API won't see the second block. We conceptually merge the blocks.

Now imagine the second was an empty block. I'd like to know -- as the developer of this library -- if that happened.

Get it?

ltratt: Ah, I see, because it would mean you'd have to think about merging in a different way?

vext01: Hrm, not sure what you mean.

Basically the user doesn't care to see partial blocks, but I do care to know if a partial block (or a complete block for that matter) is empty. I don't yet know under what circumstances that can happen so I'd like for the program to crash for now.

If at the libipt level, we had:
```
[block]
mov a, b
add b, c
[block]
<empty!>
```

then the user of our API sees:
```
[block]
mov a, b
add b, c
```

They would never see the empty block, so we need to check at the lower level.

vext01:
I think that's everything resolved then. Squash?

vext01:
You mean like that? (There was a stray negation in your comment).

I don't really mind either formulation. Yours is (gasp) one byte longer :feelsgood:

ltratt:
Oops, sorry about the extra `!`. I marginally prefer the `is_none()` but I don't feel strongly either way. Squash as you see fit (and keep that change or not as you see fit).

PR #70, opened 2018-05-14
Kill the explicit destroy in the Tracer API.

vext01:
As discussed on mattermost, this was once needed, but no more. As you can see the `destroy` methods don't really do any clean up at all.

OK?


ltratt:
It's hard to argue with having less code to deal with!

vext01:
Indeed. And this makes dealing with tracers much more manageable too!

PR #71, opened 2018-05-14
build.rs: Use an absolute dep path so consumers inherit a valid path.

vext01:
This means that anything that uses hwtracer inherits the linker path needed to find libipt. It also works transitively, so `hwtracer -> yorickrt -> your_interpreter` works.

OK?

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -126,6 +126,11 @@ fn cpu_supports_pt() -> bool {
 fn main() {
     let mut c_build = gcc::Build::new();
 
+    // We need the C_DEPS_PATH to be absolute so that our consumers inherit correct linker paths.
+    let mut c_deps_path_abs = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    c_deps_path_abs.push(C_DEPS_PATH);
+    let c_deps_str = c_deps_path_abs.display();
+
     // Check if we should build the perf_pt backend.
     if cfg!(all(target_os = "linux", target_arch = "x86_64")) {
         if feature_check("check_perf_pt.c") {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -126,6 +126,10 @@ fn cpu_supports_pt() -> bool {
 fn main() {
     let mut c_build = gcc::Build::new();
 
+    // We need the C_DEPS_PATH to be absolute so that our consumers inherit correct linker paths.
+    let mut c_deps_path_abs = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    c_deps_path_abs.push(C_DEPS_PATH);
```
ltratt: This doesn't feel right to me: `C_DEPS_PATH` should be `C_DEPS_LEAF` if it's relative to `CARGO_MANIFEST_DIR`?

build.rs
```diff
@@ -147,10 +151,10 @@ fn main() {
                 println!("cargo:rustc-env=PTXED={}/bin/ptxed", val);
             } else {
                 build_libipt();
-                c_build.include(&format!("{}/inst/include/", C_DEPS_PATH));
-                c_build.flag(&format!("-L{}/inst/lib", C_DEPS_PATH));
-                println!("cargo:rustc-link-search={}/inst/lib", C_DEPS_PATH);
-                println!("cargo:rustc-env=PTXED={}/inst/bin/ptxed", C_DEPS_PATH);
+                c_build.include(&format!("{}/inst/include/", c_deps_path_abs.display()));
```
ltratt: This feels like we want to call `display` once on `c_deps_paths_abs`.

build.rs
vext01: A path can be relative :P I've never heard of calling a directory a leaf.

If you feel strongly, how about instead `C_DEPS_DIR` instead?

build.rs
vext01: Could do yeah.

vext01:
Fixed the thing about `display()`.

build.rs
ltratt: Works for me.

vext01:
Synced branch while we are at it. OK to squash?

ltratt:
Please squash.

BTW, please don't do a sync without being asked as it messes about with the commit IDs and Github's notion of what I have and haven't seen.

vext01:
I realised afterwards, I should have left the sync as a merge commit. Sorry about that.

vext01:
Squashed. The rename is separate so that it's clear what the first change really did.

PR #72, opened 2018-05-17
Misc concurrency

vext01:
A couple of small changes needed for a larger change in yorickrt.

Merge this one first.


src/backends/perf_pt/mod.rs
```diff
@@ -310,6 +310,9 @@ impl Drop for PerfPTTrace {
     }
 }
 
+/// We promise to never dereference `Self::buf` unless we are sure what it points to is valid.
```
ltratt: I'm not sure I understand the comment or why that relates to the `unsafe impl`.

vext01: The struct is not `Send` because raw pointers are not `Send`. The reason they are not send is because we might be pointing to something thread specific.

By forcing `Send` we are saying "I know what I'm doing".

See this SO question:
https://stackoverflow.com/a/50341075/109414

[None of this stuff is very well documented]

ltratt: OK, I get it. But the comment is, if you'll forgive me, completely unclear except to someone who already understands why the `unsafe impl` is here. Can you try a substantial rephrase perhaps along the lines of "PerfPTTrace contains a raw pointer "buf", which means that Rust does not automatically derive Send. However, since we know that our implementation will never dereference "buf" unless it is safe to do so, we manually mark the struct as being Send." I wouldn't mind hinting a bit more at why we're so sure we're careful with "buf".

That said, this does me wonder if "buf" should be wrapped in its own struct (i.e. on its own) where we would do the `unsafe impl` trick. That way, if `PerfPTTrace` later gains another raw pointer, it won't accidentally be made `Send`.

vext01: Sure!

vext01:
OK, that fixes that.

I have another idea. Watch this space.

vext01:
No, I changed my mind.

Do the last two changes work for you?

ltratt:
Yep! Time to squash.

vext01:
Do you approve of my backticks? If so, I'll squash.

ltratt:
Les tiques de backs, c'est magnifique. Squashez s'il vous plait.

vext01:
fin.

PR #73, opened 2018-05-21
Function allowing consumers to discover (at compile time) which backends are available.

vext01:
Further to our discussion on backends, this small change adds a function which can be called by consumers at compile time (in a `build.rs`) giving them knowledge of which backends are available.

I've "re-exported" our `perf_pt` cfg as `hwtracer_perf_pt` so as the lessen the likelihood of a clash. For consistency, perhaps we should rename it inside hwtracer too? Thoughts.

For example usage see:
https://github.com/softdevteam/yorickrt/pull/7/commits/e0ceaffd06503b3cd17ad836892f803c260255cc

Looks good?


src/backends/mod.rs
```diff
@@ -40,5 +40,12 @@ pub mod perf_pt;
 #[cfg(perf_pt)]
 pub use self::perf_pt::PerfPTTracer;
 
+/// Call this from a consumer's build.rs to give it knowledge of which backends are available at
```
ltratt: This is a really awkward phrasing: can we simplify and shorten?

vext01:
Try this. Also, don't forget to answer the question in the PR description about renaming.

ltratt:
Agreed on the renaming. The new commit looks good too.

vext01:
Let's hold off on this. The change I'm making might render in redundant (hard to say).

vext01:
`build` is fine by me.

vext01:
OK, before I continue, let me throw a spanner in the works. As usual, threads...

This is how I reckon this would hang together with our `MetaTracer` struct. The client does something like:
```
let builder = BackendBuilder::new().perf_aux_bufsize(256);
let mt = MetaTracer::new(builder);
```

The metatracer will then use the builder to lazily make thread local tracers as and when a new thread asks to trace something.

As is stands, this means we only find out if a backend doesn't work when we call the `jit_control_point` which deeply troubles me.

I think we need to (as you suggest above) have a "probe" function (a bit like `match()` in OpenBSD drivers) which decides if a backend should work on the current system.

I think we can combine that with a list of backends to try in succession. The first one which should work is used.

Sounds good?

[one more question in the stack after this one]

vext01:
Tracers are bound to a single thread due to the way perf works.

ltratt:
Why does that matter to the PT struct?

vext01:
I'm not sure what you mean. Each `PerfPTTracer` instance is bound to one thread.

ltratt:
That might not be the easiest API. Maybe we could say something like "PT has a method `trace_pt` which returns the appropriate PTTracer struct for the current thread"?

vext01:
I'm not saying no, but isn't that what we are doing with this builder API?

vext01:
> We seem to be putting the burden of creating one tracer struct per thread onto the client

No, that's not true. We ask the client to create a *builder* which informs the metatracer how to automatically make new tracers as it requires them.

ltratt:
Yes, true, but as you pointed out that means `MetaTracer` now handles the errors. Maybe we need an intermediate option in between a builder and a Tracer object? That would allow clients to catch errors, and `MetaTracer` to ignore builder-type errors.

vext01:
> Yes, true, but as you pointed out that means MetaTracer now handles the errors.

Indeed. I've been thinking the same, and here's another reason why: I don't want the backend detection routines to run each and every time a new thread local tracer is instantiated. All of the threads will use the same tracer type (i hope!).

Let me have a think. I'll propose a new API before doing any more coding.

ltratt:
I would suggest something vaguely like:

```rust

struct PTTracerBuilder { }
struct PTTracer { }
struct PTTracerThread { }
```

where users start with a `PTTracerBuilder`, set options on it; and at some point they say `build` to get them a `PTTracer` struct. That only really has 1 function `tracer_for_thread()` which returns the `PTTracerThread` struct specialised to the current thread. So users would pass `PTTracer` to `MetaTracer` and `MetaTracer` (internally) would call `tracer_for_thread()`. That means the validation costs are only paid when you call `build`.

ltratt:
If you don't check what tracer you got, then you probably shouldn't be setting options. But we want to handle the default case nicely with something simple like `PCTraceBuilder::new().build()`.

vext01:
I think you misunderstood me. I thought you wanted it so that a list of possibly usable backends were checked in turn. In which case you'd need to know which one the system chose...

ltratt:
I refer you to my `backend: Option` comment a little way above for my suggestion there.

vext01:
Oh, so you want it so that a list is searched only in the default case?

vext01:
I'm confused, you had a generic builder before, but later we have builders specific to the different backends.

ltratt:
I think you can merge these concepts together pretty easily. I don't think there's anything complex here.

vext01:
I might understand. I'll come back to this tomorrow morning with a fresh head and see if it makes more sense.

vext01:
Hi Laurie,

Here's another *rough* prototype of what I think you described yesterday. I changed some type names to make more sense in my head. Please comment.

It works like this:

## Make thread tracers for a automatically chosen default backend.
```
        let tracer = TracerBuilder::new().build().unwrap();
        let tt1 = tracer.thread_tracer();
        let tt2 = tracer.thread_tracer();
        t1.start_tracing();
        ...
```

## Make thread tracers for a manually chosen backend with custom config.
```
        let tracer = TracerBuilder::new().backend(BackendKind::PerfPT).perf_aux_bufsize(8192).build().unwrap();
        let tt1 = tracer.thread_tracer();
        let tt2 = tracer.thread_tracer();
        ...
```

## Make thread tracers for a automatically chosen default backend, but with a custom config if `PerfPT` is chosen:

```
       let mut bldr = TracerBuilder::new();    
       bldr = match bldr.backend_kind() {    
            BackendKind::Dummy => bldr,    
            BackendKind::PerfPT => bldr.perf_aux_bufsize(8192),    
        };
        let tracer = bldr.build().unwrap();
        ...
```

## Notes

 * Killed `target_tid` as a config variable, as under the new API the target TID is always the TID under which `thread_tracer()` is called (I think... not hard to bring back if I was wrong).

Looks like we are on the right track? If so I will get this PR ready.

ltratt:
Yes, this looks like a good route to me. Please go ahead.

[Though that last example would surely be simpler as:
```rust
       let mut bldr = TracerBuilder::new();    
       if let BackendKind::PerfPT = bldr.backend_kind() { 
           bldr.perf_aux_bufsize(8192);
       }
       let tracer = bldr.build().unwrap();
```
not that it makes any difference to the main body of code.]

vext01:
> Yes, this looks like a good route to me. Please go ahead.

Cool!

> not that it makes any difference to the main body of code

Yeah, just something I hacked up for example purposes only.

ltratt:
Feel free to clean up the new commits (I assumed you'd done that before pushing), but don't squash into the old commits.

vext01:
So where do old commits start?

We've not commented on any actual code yet, why don't we just squash it all into one? Or close this and I'll make a new PR?

vext01:
Here we go then. I imagine you will have many questions. Shoot.

src/backends/dummy.rs
```diff
@@ -107,38 +132,52 @@ impl Iterator for DummyBlockIterator {
...
     }
+
+    #[test]
+    fn test_unapplicable_config() {
+        let res = TracerBuilder::new()
+                                .backend(BackendKind::Dummy)
+                                .perf_aux_bufsize(1000) // Invalid.
+                                .build();
+        let expect = "perf_aux_bufsize config attribute does not apply to DummyThreadTracer";
+        match res {
```
ltratt: This could just be `assert_eq!(res.unwrap_err(), expect);`.

src/backends/mod.rs
```diff
@@ -35,10 +35,205 @@
...
 pub use self::perf_pt::PerfPTTracer;
 
+#[derive(Debug, PartialEq)]
+pub enum BackendKind {
+    Dummy,
+    PerfPT,
+}
+
+/// Generic configuration interface for all backends.
+/// A `None` field indicates a default value and attributes which don't apply to a given backend
```
ltratt: "If an attribute is set to `None`, a default value will be filled in later by <whatever component does that>."

src/backends/dummy.rs
vext01: Nice

src/backends/mod.rs
```diff
@@ -35,10 +35,205 @@
...
+pub struct BackendConfig {
+    perf_aux_bufsize: Option<usize>,
+    perf_data_bufsize: Option<usize>,
+    new_trace_bufsize: Option<usize>,
+}
+
+/// A builder interface for instantiating backends.
+pub struct TracerBuilder {
+    user_backend: Option<BackendKind>,  // User-requested backend choice.
+    auto_backend: BackendKind,          // Auto-chosen backend if `user_backend` is `None`.
```
ltratt: This attribute seems pointless? Surely it should be the job of some later phase to say "if user_backend is None: self.default_platform_backend()"?

src/backends/mod.rs
```diff
@@ -35,10 +35,205 @@
...
+pub enum BackendKind {
+    Dummy,
+    PerfPT,
+}
+
+/// Generic configuration interface for all backends.
+/// A `None` field indicates a default value and attributes which don't apply to a given backend
+/// are checked by the `TracerBuilder` when `build()` is called.
+#[derive(Debug)]
+pub struct BackendConfig {
```
ltratt: What's the use in separating this struct from `TracerBuilder`?

src/backends/mod.rs
```diff
@@ -35,10 +35,205 @@
...
+        for kind in tr_kinds {
+            if Self::match_platform(&kind).is_ok() {
+                return kind;
+            }
+        }
+        // The Dummy backend should always be usable.
+        unreachable!();
+    }
+
+    /// Returns `Ok` if the tracer backend should work on the current hardware/OS.
```
ltratt: This is a weird comment given the function name. Should the function be renamed?

src/backends/mod.rs
```diff
@@ -35,10 +35,205 @@
...
+        match kind {
+            BackendKind::Dummy => Ok(()),
+            BackendKind::PerfPT => {
+                #[cfg(not(perf_pt))]
+                return Err(HWTracerError::BackendUnavailable(BackendKind::PerfPT));
+                #[cfg(perf_pt)] {
+                    if !Self::pt_supported() {
+                        return Err(HWTracerError::NoHWSupport("Intel PT not supported by CPU".into()));
+                    }
+                    return Ok(());
```
ltratt: Can just be `Ok(())`.

src/backends/mod.rs
vext01: It get's passed into `from_backend_config()` of the `Tracer` trait, which no longer needs the `backend` attribute.

src/backends/mod.rs
vext01: We have to populate it early so that the user can set a custom configuration on an auto-chosen backend. See "Make thread tracers for a automatically chosen default backend, but with a custom config if PerfPT is chosen" above.

src/backends/perf_pt/mod.rs
```diff
@@ -323,65 +321,92 @@ impl Drop for PerfPTTrace {
...
+impl PerfPTTracer {
+    fn check_perf_perms() -> Result<(), HWTracerError> {
+        if unsafe { geteuid() } == 0 {
+            // Root can always trace.
+            return Ok(());
+        }
+
+        let mut f = File::open(&PERF_PERMS_PATH)?;
+        let mut buf = String::new();
+        f.read_to_string(&mut buf)?;
```
ltratt: I think you can replace this with https://doc.rust-lang.org/std/fs/fn.read_to_string.html

src/backends/mod.rs
vext01: `matches_platform` would make more sense?

src/backends/mod.rs
vext01: Maybe. Conditional compilation often gets in the way of expression evaluation. I've been meaning to make a bug report. Will try.

src/backends/perf_pt/mod.rs
```diff
@@ -323,65 +321,92 @@ impl Drop for PerfPTTrace {
...
+        }
+        if config.perf_aux_bufsize.is_some() {
+            ppt_config.aux_bufsize = config.perf_aux_bufsize.unwrap();
+        }
+        if config.new_trace_bufsize.is_some() {
+            ppt_config.new_trace_bufsize = config.new_trace_bufsize.unwrap();
+        }
+
+        // Check for inavlid configuration.
+        fn power_of_2(v: size_t) -> bool {
```
ltratt: Isn't `size_t` unsigned in Rust? If so the `v <= 0` check is unneeded.

src/lib.rs
```diff
@@ -91,17 +93,21 @@ pub trait Trace: Debug + Send {
 }
 
 /// The interface offered by all tracer types.
-///
-/// It is the job of the consumer of this library to decide which specific tracer type to
-/// instantiate, and to set it up using its type-specific configuration methods.
-pub trait Tracer {
+pub trait Tracer: Send + Sync {
```
ltratt: Do you need the explicit `Send` and `Sync`?

src/backends/mod.rs
vext01: Yeah, this one actually works.

src/backends/mod.rs
ltratt: Is it worth the hassle of two structs?

src/backends/mod.rs
vext01: Here's an example of the kind of stuff I've noticed:
```
fn num() -> usize {
    #[cfg(test)]
    return 1;
    #[cfg(not(test))]
    2
}
```

Without the `return` this fails to compile. Yet the second `#[cfg]` and the following line should simply disappear. Weird.

src/backends/perf_pt/mod.rs
vext01: Indeed!

src/backends/mod.rs
ltratt: It's not really surprising: it's a sort-of limitation of the parsing. You might think you could write:

```rust
fn num() -> usize {
    #[cfg(test)]
    1
    #[cfg(not(test))]
    2
}
```

but (I imagine) that's unparseable even though "semantically" it makes sense.

src/backends/perf_pt/mod.rs
vext01: Well I never!

```
type size_t = usize;
```

But that doesn't match C! And this is in the libc crate. Suspicious...

vext01: Ah no wait, in C you have `ssize_t`. Yes, you are right.

src/lib.rs
ltratt: No: they're auto-traits.

src/backends/mod.rs
vext01: Up to you. There's not really any cost in passing the redundant field into `from_backend_config`. The struct is short-lived and that field would be ignored if we did pass it in.

src/backends/mod.rs
vext01: Yes, that's exactly what I thought I could write. Blame C.

```
long long int num() {
    #ifdef TEST
    return 1;
    #else
    return 2;
}
```

Thanks to the preprocessor the parser doesn't even see one arm of the compile time constant. I assumed Rust would do the same, but now I see why it cannot.

src/lib.rs
vext01: Well yes, but this decision would move the error.

Without the annotation you would get an error near a spawn closure.

With you would get an error saying that the backend doesn't conform to our API.

src/backends/mod.rs
ltratt: I think I'd suggest one struct to keep things simple.

src/backends/perf_pt/mod.rs
```diff
@@ -323,65 +321,92 @@ impl Drop for PerfPTTrace {
...
-        self
+        Ok(())
     }
+}
+
+impl Tracer for PerfPTTracer {
+    fn from_backend_config(config: &BackendConfig) -> Result<Box<Tracer>, HWTracerError> where Self: Sized {
+        let mut ppt_config = PerfPTConf::default();
+
+        if config.perf_data_bufsize.is_some() {
```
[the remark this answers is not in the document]
ltratt: Good point.

src/lib.rs
ltratt: I don't get it. They trait is `Send + Sync` automatically (I think), so how can removing these change anything?

vext01: A pub trait cannot be `Send + Sync` automatically, only it's implementers.

vext01: That's the spirit!

src/backends/dummy.rs
vext01: Huh, this is odd:
```
o method named `unwrap_err` found for type `std::result::Result<...>
```

And the functionality isn't in a trait...

vext01: [Note that this would enforce this on our pubic API]

ltratt: No, we don't want to do that. Maybe it should be `assert_eq!(res.unwrap_err().to_string(), expect);`

src/backends/mod.rs
vext01: Hrm, I'm looking at the struct inlined and thinking "this looks worse than before". We've got the management fields mixed in with the config knobs. Could I convince you to reconsider?

```
 /// A builder interface for instantiating backends.    
  pub struct TracerBuilder {    
      user_backend: Option<BackendKind>,  // User-requested backend choice.    
      auto_backend: BackendKind,          // Auto-chosen backend if `user_backend` is `None`.
      config: BackendConfig,    
      /// Generic backend configuration knobs.    
      /// If a field is `None` at `build()` time then the backend will select a default value. Ant
      /// attributes which don't apply to a given backend are also checked.          
      perf_aux_bufsize: Option<usize>,                                         
      perf_data_bufsize: Option<usize>,                     
      new_trace_bufsize: Option<usize>,    
  }
```

ltratt: It would surely be:

```
  pub struct TracerBuilder {    
      user_backend: Option<BackendKind>,  // User-requested backend choice.         
      perf_aux_bufsize: Option<usize>,                                         
      perf_data_bufsize: Option<usize>,                     
      new_trace_bufsize: Option<usize>,    
  }
```

src/backends/dummy.rs
vext01: Well that would kill the need for `PartialEq` but not the need for `Debug`. You need to avoid using `unwrap_err` if you don't want `Debug`.

vext01:
```
impl<T, E> Result<T, E> where T: Debug {  // <--- Debug!
    pub fn unwrap_err(self) -> E
```

ltratt: I have no problem with `Debug`, but it's up to you as to whether you think the cost is worth it.

vext01: I don't think there's any (performance) cost. It's just we force people implementing backends to implement `Debug`, which isn't strictly required.

vext01:
> So one over-arching question I have is that we seem to have put backend specific functions (e.g. the AUX buffer size) into the generic interface. That means the public API will be the union of every possible configuration.

Yes. This is why I asked you if I should continue 😆

> Maybe the backend should be dynamically dispatched via a Box or similar?

I don't know what you mean. Can you show a (compiling) example?

vext01:
> their code just wouldn't be able to compile using a non-applicable option

 I *think* 💭

vext01:
Actually, I don't see how that could work.

You'd need something like the property interfaces you see all over java.

```
pt_builder.set_option(ConfigOption::AUX_BUFSIZE, ConfigVal::Usize{1024})
```

ltratt:
You'd need to have an enum something like:

```
enum BackendConfig {
    Dummy,
    PT(PTConfig)
}

fn backend_config(&self) -> &mut BackendConfig {
    ...
}
```

and then you could use it something like:

```rust
match builder.backend_config() {
    PT(ptconf) => {
        // set options on ptconf
    }
}
```

I'm not saying that's perfect, but hopefully you get the rough idea.

vext01:
I hadn't thought of using an enum before. Seems like it could work, but I'd have to try it out.

If it works, that's a better solution to one of my stackoverflow questions.

I'll start on this first thing tomorrow.

vext01:
So, something like this?

[There is a simplification which I can apply if this is looking OK: We can directly use `PerfPTConf` from `backends/perf_pt/mod.rs` meaning that we can remove `PerfPTBackendConfig`]

Good starting point for reading would be these tests:

https://github.com/softdevteam/hwtracer/blob/669960cc134a237a70352ef76fb9533023328014/src/backends/mod.rs#L212-L236

ltratt:
For:

```rust
     if let BackendConfig::PerfPT(ref mut ppt_conf) = bldr.config { 
         ppt_conf.aux_bufsize = Some(3); 
     } else { 
         unreachable!(); 
     } 
     match bldr.build() { 
         Err(HWTracerError::BadConfig(s)) => { 
             assert_eq!(s, "aux_bufsize must be a positive power of 2"); 
         }, 
         _ => panic!(), 
    } 
```

why not use `if` for both or `match` for both?

vext01:
We could do yeah. But is the API on the right lines? I'm still not certain. What do you think?

ltratt:
This looks pretty much perfect. I would suggest that `config` is a private field and users access it via a `config` function (just in case we want to change the representation later), but otherwise I think this looks good.

vext01:
> This looks pretty much perfect.

I'm both surprised an relieved. I really didn't like the destructuring step, especially with necessity for `ref mut`. I think we will need to show some examples because I wouldn't have found that obvious.

> I would suggest that config is a private field and users access it via a config function

OK, I'll push a commit for that shortly.

I'm also going to un-pub the constructors for the concrete backends so that clients can't bypass the platform checks in the builder.

vext01:
The last commit needs discussion. Basically we can stick the struct that was private to `backends/perf_pt/mod.rs` directly into our new `BackendConfig` enum thus avoiding some duplication and making it easier for the user (there's no `Option` inside the configs any more, since the user simply overwrites the already present defaults).

What do you think of that?

src/backends/mod.rs
```diff
@@ -109,48 +109,56 @@ impl BackendKind {
...
+    pub aux_bufsize: size_t,
+    // The initial trace buffer size (in bytes) for new traces.
+    pub new_trace_bufsize: size_t,
 }
 
-impl PerfPTBackendConfig {
+impl Default for PerfPTConfig {
+    fn default() -> Self {
+        Self {
+            data_bufsize: 64,
```
ltratt: I think these should be `CONSTANTS` at the top of the file.

ltratt:
It doesn't have any effect on the external API AFAICS, and it seems like a reasonable simplification, so I'm OK with it.

vext01:
The only effect it has on the external API is that the user will see FFI types directly. I think that's OK.

The last commit addresses the last batch of comments.

While we are in this code, every time I see `new_trace_bufsize` I feel like `initial_trace_bufsize` would capture the meaning better. It implies that the actual size may change. What do you think?

ltratt:
Agreed on the new -> initial name change.

ltratt:
Works for me. Time to squash?

vext01:
Splat.

PR #74, opened 2018-06-29
Kill dead code.

vext01:
There's dome dead code in hwtracer which causes a linking error for `--release` builds. This change kills the dead code.

The code in question is a leftover from a time where I was going to allow tracing a PID other than yourself. We since hard-coded the PID to the current thread in C code.

OK?


vext01:
Test failures... bear with me.

vext01:
What happened here was the Rust errors we were looking for changed:
 * `dyn` appears
 * An error message was re-worded.

If fixed this and made the thing we look for a little more generic, however, it wouldn't have protected us from the latter point above.

Squash?

ltratt:
Yep, please squash.

vext01:
splat

PR #77, opened 2018-10-30
Build fixes

ltratt:
This PR makes things build on OpenBSD and updates several dependencies. Mostly these are easy but two (gcc -> cc; and tempfile 2.* to 3.*) require minor code changes.

This passes all tests on bencher8 (except `backends::perf_pt::tests::test_block_iterator5` which it skips, presumably intentionally?).

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -201,7 +201,7 @@ impl<'t> PerfPTBlockIterator<'t> {
             Err(cerr)?
         }
 
-        vdso_tempfile.sync_all()?;
+        vdso_tempfile.as_file().sync_all()?;
         self.decoder = decoder;
         self.vdso_tempfile = Some(vdso_tempfile);
         Ok(())
@@ -532,9 +532,9 @@ mod tests {
     // Given a trace, use ptxed to get a vector of block start vaddrs.
     fn get_expected_blocks(trace: &Box<Trace>) -> Vec<Block> {
         // Write the trace out to a temp file so ptxed can decode it.
-        let mut fh = NamedTempFile::new().unwrap();
-        trace.to_file(&mut fh);
-        let (args, _vdso_tempfile) = self_ptxed_args(fh.path().to_str().unwrap());
+        let mut tmpf = NamedTempFile::new().unwrap();
+        trace.to_file(&mut tmpf.as_file_mut());
+        let (args, _vdso_tempfile) = self_ptxed_args(tmpf.path().to_str().unwrap());
 
         let out = Command::new(env!("PTXED"))
                           .args(&args)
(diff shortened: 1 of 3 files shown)


src/backends/perf_pt/mod.rs
```diff
@@ -532,9 +532,9 @@ mod tests {
     // Given a trace, use ptxed to get a vector of block start vaddrs.
     fn get_expected_blocks(trace: &Box<Trace>) -> Vec<Block> {
         // Write the trace out to a temp file so ptxed can decode it.
-        let mut fh = NamedTempFile::new().unwrap();
```
vext01: Why the renaming of `fh` to `tmpf`? If we leave it as `fh` the change touches fewer lines (OK, only one, but still).

src/backends/perf_pt/mod.rs
```diff
@@ -624,12 +624,12 @@ mod tests {
...
         // Make the trace and write it to a file.
-        let mut fh = NamedTempFile::new().unwrap();
-        trace.to_file(&mut fh);
-        fh.sync_all().unwrap();
+        let mut tmpf = NamedTempFile::new().unwrap();
+        trace.to_file(&mut tmpf.as_file_mut());
+        tmpf.as_file().sync_all().unwrap();
 
         // Check the resulting file makes sense.
-        let file = File::open(fh.path().to_str().unwrap()).unwrap();
```
vext01: Similarly, this change would not be necessary if we left the variable name as-is.

vext01:
`backends::perf_pt::tests::test_block_iterator5` has an `#[ignore]` because it's  a slow test. To run it you have to pass `--ignored` (IIRC) to the `cargo test` invocation.

https://doc.rust-lang.org/book/second-edition/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested

Other than my small comments, this all looks OK to me.

src/backends/perf_pt/mod.rs
ltratt: Because it's no longer a file handle...

vext01: > Because it's no longer a file handle

Well, it was a handle on a temporary file before, and that's what it still is.

I don't care greatly, but I think it's good practice to touch as few lines as possible unless there's a good reason.

Up to you...

ltratt:
[For reference it's `cargo test -- --ignored`]

src/backends/perf_pt/mod.rs
vext01: I feel like you are splitting hairs, but as I said, it's up to you. So are you ready to merge?

ltratt: Yep!

PR #78, opened 2019-09-18
Fix warnings and errors

ptersilie:
This PR fixes some warnings and errors resulting from changes in some dependencies and rustc.


vext01:
LGTM, but I think the second commit message could be expanded to explain what we've done.

ptersilie:
Should I also squash the last commit while I'm doing that?

vext01:
I think these make sense as separate commits in this instance.

ptersilie:
Commit message updated.

PR #81, opened 2020-04-06
Rename start_vaddr and len

ptersilie:
This will break our yk repo, so I'll have to raise a "companion" PR with this to fix that. However, there's currently some other stuff going on on the yk repo, so let's wait until that's finished. However, we can start the review in the meantime.

The current name for these variables suggests that they give the range
of the block, which would allow to calculate the next block by adding
together `start_vaddr` and `len`. This is not the case though, since len
is calculated from `block.end_ip` which returns the address of the last
instruction in this block. This means that a block with a single
instruction would have a `len` of 0, which is unintuitive.

This commit renames the variables to `first_instr` and `last_instr` and
changes the latter to return the addresses of the last instruction of
the block (instead of returning its length).

### DIFF
--- a/src/backends/perf_pt/decode.c
+++ b/src/backends/perf_pt/decode.c
@@ -170,27 +170,27 @@ clean:
 }
 
 /*
- * Updates `*addr` and `*len` with the start address and the length of the next
- * block in the instruction stream.
+ * Updates `*first_instr` and `*last_instr` with the address of the first and last
+ * instructions of the next block in the instruction stream.
  *
- * A start address of 0 indicates that the end of the instruction stream has
- * been reached.
+ * If first instruction address is 0, this indicates that the end of
+ * the instruction stream has been reached.
  *
  * `*decoder_status` will be updated with the new decoder status after the operation.
  *
- * Returns true on success or false otherwise. Upon failure, `*addr` and `*len`
- * are undefined.
+ * Returns true on success or false otherwise. Upon failure, `*first_instr` and
+ * `*last_instr` are undefined.
  */
 bool
 perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
-                   uint64_t *addr, uint64_t *len, struct perf_pt_cerror *err) {
+        uint64_t *first_instr, uint64_t *last_instr, struct perf_pt_cerror *err) {
     // If there are events pending, look at those first.
     if (handle_events(decoder, decoder_status, err) != true) {
         // handle_events will have already called perf_pt_set_err().
         return false;
     } else if (*decoder_status & pts_eos) {
         // End of stream.
-        *addr = 0;
+        *first_instr = 0;
         return true;
     }
     if ((*decoder_status != 0) && (*decoder_status != pts_ip_suppressed)) {
@@ -204,14 +204,14 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
     struct pt_block block;
     block.iclass = ptic_other;
     bool first_block = true;
-    *len = 0;
+    *last_instr = 0;
     while (!block_is_terminated(&block)) {
         if (handle_events(decoder, decoder_status, err) != true) {
             // handle_events will have already called perf_pt_set_err().
             return false;
         } else if (*decoder_status & pts_eos) {
             // End of stream.
-            *addr = 0;
+            *first_instr = 0;
             return true;
         }
         // It's possible at this point that we get notified of an event in the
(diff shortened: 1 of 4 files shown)


src/backends/perf_pt/decode.c
```diff
@@ -170,27 +170,27 @@ perf_pt_init_block_decoder(void *buf, uint64_t len, int vdso_fd, char *vdso_file
...
 
 /*
- * Updates `*addr` and `*len` with the start address and the length of the next
- * block in the instruction stream.
+ * Updates `*first_instr` and `*last_instr` with the address of the first and last
+ * instructions of the next block in the instruction stream.
  *
- * A start address of 0 indicates that the end of the instruction stream has
- * been reached.
+ * An instruction address of 0 indicates that the end of the instruction stream
```
vext01, line 176: Which address is set to zero? The `first_instr`?

src/backends/perf_pt/decode.c
```diff
@@ -246,14 +246,15 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
         }
 
         if (first_block) {
-            // The block start address we report back to the user.
-            *addr = block.ip;
+            // The address of the blocks first instruction that we report back
```
vext01, line 249: block's

src/backends/perf_pt/decode.c
```diff
@@ -246,14 +246,15 @@ perf_pt_next_block(struct pt_block_decoder *decoder, int *decoder_status,
...
+            // to the user.
+            *first_instr = block.ip;
             first_block = false;
         }
 
-        // Update the length.
-        *len += block.end_ip - block.ip;
     }
+    // The address of the blocks last instruction.
```
vext01, line 256: block's again.

src/backends/perf_pt/decode.c
ptersilie: The first, but actually I think we should set both to zero.

src/backends/perf_pt/mod.rs
```diff
@@ -231,21 +231,21 @@ impl<'t> Iterator for PerfPTBlockIterator<'t> {
             }
         }
 
-        let mut addr = 0;
-        let mut len = 0;
+        let mut first = 0;
```
vext01, line 234: Shall we go for `first_instr` and `last_instr` for consistency?

src/lib.rs
```diff
@@ -59,21 +59,26 @@ use std::fs::File;
 /// Information about a basic block.
 #[derive(Debug, Eq, PartialEq)]
 pub struct Block {
-    start_vaddr: u64,
-    len: u64,
+    /// Virtual address of the first instruction in this block
```
vext01, line 62: Add full stops.

src/backends/perf_pt/decode.c
vext01: That's fine, but let's document it.

vext01:
Just very minor comments.

ptersilie:
Addressed comments.

src/backends/perf_pt/decode.c
```diff
@@ -173,8 +173,8 @@ perf_pt_init_block_decoder(void *buf, uint64_t len, int vdso_fd, char *vdso_file
  * Updates `*first_instr` and `*last_instr` with the address of the first and last
  * instructions of the next block in the instruction stream.
  *
- * An instruction address of 0 indicates that the end of the instruction stream
- * has been reached.
+ * If the returned instruction addresses are 0, this indicates that the end of
```
vext01, line 176: A user might have the question: do I have to check both are zero, or is it sufficient to check only one?

ptersilie: Fixed.

vext01:
LGTM. Please squash.

ptersilie:
Squashed. Though let's not merge this until I have the yk PR at the ready. I'll let you know.

vext01:
ok

ptersilie:
This is ready for merging now.

ptersilie:
Sorry, forgot about the tests. Let me know when to squash.

vext01:
Please squash.

ptersilie:
Squashed.

ptersilie:
I don't think this repo is in bors.

vext01:
I take it you checked it `cargo fmt`s and tests work?

ptersilie:
Tests work. `cargo fmt` has never been run on this repo it seems.

vext01:
> Tests work. cargo fmt has never been run on this repo it seems.

OK, I'm going to add this to CI tomorrow then.

PR #87, opened 2020-07-10
Simplify tests

vext01:
Realised that we could make the code a lot simpler with our `phdrs` crate.

See commit messages.


vext01:
Ready. No squash needed.

PR #88, opened 2020-07-15
Fix rebuilds.

vext01:
This almost fixes the rebuilding issue we've been having.

See the second commit message for details on why I say "almost". I've spent far too long trying to work this out, and I feel my sanity slipping away with every keystroke.

At least I've identified which paths cause the problem.

It seems to be that rerun-if-changed doesn't deal well with files that didn't initially exist (that are generated), but beyond that I'm not really sure what's happening. By all means, have a play around and see if you can figure it out...

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -5,10 +5,11 @@ set -e
 export CARGO_HOME="`pwd`/.cargo"
 export RUSTUP_HOME="`pwd`/.rustup"
 
+# Install no toolchain initially to ensure toolchain rollback.
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
-sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain nightly -y --no-modify-path
-
+sh rustup.sh --default-host x86_64-unknown-linux-gnu --default-toolchain none -y --no-modify-path
 export PATH=`pwd`/.cargo/bin/:$PATH
+rustup install nightly
 
 cargo fmt --all -- --check
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,4 @@ phdrs = { git = "https://github.com/softdevteam/phdrs" }
 
 [build-dependencies]
 cc = "1.0.57"
+rerun_except = "0.1.2"
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,6 @@
 use cc;
 use core::arch::x86_64::__cpuid_count;
+use rerun_except::rerun_except;
 use std::env;
 use std::fs;
 use std::os::unix::fs as unix_fs;
@@ -149,15 +150,13 @@ fn main() {
 
     // Additional circumstances under which to re-run this build.rs.
     println!("cargo:rerun-if-env-changed=IPT_PATH");
-    println!("cargo:rerun-if-changed=src/util");
-    println!("cargo:rerun-if-changed={}", c_deps_dir_s);
-    println!("cargo:rerun-if-changed=src/backends/perf_pt");
-    println!(
-        "cargo:rerun-if-changed={}/processor-trace/libipt/src/pt_cpu.c",
-        c_deps_dir_s
-    );
-    println!(
-        "cargo:rerun-if-changed={}/processor-trace/libipt/src/posix/pt_cpuid.c",
-        c_deps_dir_s
-    );
+    rerun_except(&[
+        "README.md",
+        "deny.toml",
+        "LICENSE-*",
+        "COPYRIGHT",
+        "bors.toml",
+        ".buildbot.sh",
+    ])
+    .unwrap();
 }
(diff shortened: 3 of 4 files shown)


ltratt:
Would this not be safer with rerun_except, rather than manually specifying files to recompile?

vext01:
I'm not sure it'd help, would it?

ltratt:
It won't solve the immediate problem you have, but it's designed to solve the longer term problem. At the moment, AFAICS, you're specifying which files *should* trigger a rebuild. What happens if you later add a file which should trigger a rebuild and forget to add it to the file? That's what rerun_except solves: you specify only the files which *shouldn't* trigger a rebuild. New files (unless excluded by a glob) thus automatically trigger rebuilds which might be inefficient but is safe.

vext01:
I agree that this would be useful, but I think it's only semi-related and perhaps for a separate PR

ltratt:
It seems pretty related to me, but it's your call.

vext01:
Well, I can try, but I'm not certain it's going to help this particular issue.

I'll give it a go...

vext01:
I'm not sure rerun-except can help here, as files under `OUT_DIR` are not considered by default. As far as I can see, rerun_except can only be used to remove a glob from Cargo's default file set?

Here's a tiny example that better explains the issue:
```
use std::env;
use std::fs::File;
use std::path::PathBuf;

fn main() {
    let out_dir = env::var("OUT_DIR").unwrap();

    let turnips = [&out_dir, "turnips"].iter().collect::<PathBuf>();
    File::create(&turnips).unwrap();
}
```

Suppose we want a rebuild if the `turnips` file changes.

The build script makes a file that wouldn't be considered by the default configuration, as confirmed by:
```
$ cargo clean
$ cargo check
   Compiling xxx v0.1.0 (/home/edd/source/xxx)
    Finished dev [unoptimized + debuginfo] target(s) in 0.54s
$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
$ find . -name 'turnips'
./target/debug/build/xxx-c4af9b9f27c0c3b3/out/turnips
$ find . -name 'turnips' | xargs touch -m
$ cargo check                            
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
```

The project was not rebuilt, as the files under `OUT_DIR` are not considered by default. I think that rules out rerun_except, but let me know if I'm wrong.

We might think we can fix that with a `rerun-if-changed`:
```
use std::env;
use std::fs::File;
use std::path::PathBuf;

fn main() {
    let out_dir = env::var("OUT_DIR").unwrap();

    let turnips = [&out_dir, "turnips"].iter().collect::<PathBuf>();
    File::create(&turnips).unwrap();

    println!("cargo:rerun-if-changed={}", turnips.to_str().unwrap());
}
```

Let's try it:
```
$ cargo check
   Compiling xxx v0.1.0 (/home/edd/source/xxx)
    Finished dev [unoptimized + debuginfo] target(s) in 0.55s
$ cargo check
   Compiling xxx v0.1.0 (/home/edd/source/xxx)
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
```

It rebuilt unnecessarily the second time...

And I've just noticed that this behaviour is non-deterministic:
```
$ cargo clean
$ cargo check
   Compiling xxx v0.1.0 (/home/edd/source/xxx)
    Finished dev [unoptimized + debuginfo] target(s) in 0.55s
$ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
```

This time the second invocation *didn't* rebuild.

I also tried moving the rerun-if-changed line before creation of the file, but this makes no difference.

So all in all, I'm confused. Can you repro?

ltratt:
> ${v${MODPY_FLAVOR} \

vext01:
I pushed commits to finally fix this, but GH hasn't registered the new commits yet.

https://www.githubstatus.com/

vext01:
Splat.

vext01:
The classic:
```
error: the 'rustfmt' component which provides the command 'cargo-fmt' is not available for the 'nightly-x86_64-unknown-linux-gnu' toolchain
```

Shall I add the "rollback" hack onto this branch?

vext01:
I'd like to leave that commit separate.

ltratt:
I think you're right.

bors r+

.buildbot.sh
```diff
@@ -5,10 +5,11 @@ set -e
 export CARGO_HOME="`pwd`/.cargo"
 export RUSTUP_HOME="`pwd`/.rustup"
 
+# Install no toolchain initially ensure toolchain rollback.
```
vext01, line 8: damnit typo

bors r-

vext01:
I'm sure that change won't be controversial :)

ltratt:
Please squash.

vext01:
splat.

PR #89, opened 2020-09-03
Fix compilation error in tests.

vext01:
Another compilation error that somehow slipped CI. I don't get it...

Did we perhaps manually merge in a time where CI was down?

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -521,7 +521,7 @@ mod tests {
             };
 
             for hdr in obj.iter_phdrs() {
-                if hdr.type_() != PT_LOAD || hdr.flags() & PF_X == 0 {
+                if hdr.type_() != PT_LOAD || hdr.flags() & PF_X.0 == 0 {
                     continue; // Only look at loadable and executable segments.
                 }
 


ptersilie:
I'm as baffled as you are.

bjorn3:
That line was introduced in https://github.com/softdevteam/hwtracer/pull/87, which passed CI. I even checked that it ran `cargo test` and that tests for the `perf_pt` backend ran.

vext01:
Odd.... Let's keep an eye on this.

bjorn3:
I found the culprit. This repo doesn't have a `Cargo.lock` checked in and at the same time uses a git dependency on https://github.com/softdevteam/phdrs without pinning a commit. This meant that when https://github.com/softdevteam/phdrs/pull/5 was merged, the next CI run on this repo picked it up. That PR was a breaking change, which meant that CI on this repo broke. You should either check in `Cargo.lock` or pin the git dependency to a specific commit. (`rev =` I think)

ltratt:
Yes, `rev` in `Cargo.toml` also works: IMHO that's the preferred route, because it's robust to `Cargo.lock` changes.

vext01:
> the next CI run on this repo picked it up

I don't get this part. The CI destroys the build directory between builds:

```
buildbot_sh_factory.addStep(steps.Git(repourl=get_repourl,
                                      mode='full', method='clobber', # Start fresh each time.
                                      shallow=True))
```

http://docs.buildbot.net/current/manual/configuration/buildsteps.html#git

bjorn3:
`Cargo.lock` has to be regenerated on every CI run. Cargo chooses the HEAD commit in https://github.com/softdevteam/phdrs for the `phdrs` dependency when it needs to generate `Cargo.lock` as there is no `rev` specification.

PR #93, opened 2020-12-04
Stop tracing upon drop.

ptersilie:
It's possible for PerfPTThreadTracer to be dropped while it is still
tracing, leaving the PT tracer running. This can cause issues with other
threads wanting to trace. Calling stop_tracing on the PT tracer when the
hardware tracer is dropped fixes this.

### DIFF
--- a/src/backends/perf_pt/mod.rs
+++ b/src/backends/perf_pt/mod.rs
@@ -403,6 +403,15 @@ impl Default for PerfPTThreadTracer {
     }
 }
 
+impl Drop for PerfPTThreadTracer {
+    fn drop(&mut self) {
+        if self.state == TracerState::Started {
+            // If we haven't stopped the tracer already, stop it now.
+            self.stop_tracing().unwrap();
+        }
+    }
+}
+
 impl ThreadTracer for PerfPTThreadTracer {
     fn start_tracing(&mut self) -> Result<(), HWTracerError> {
         if self.state == TracerState::Started {


ptersilie:
@vext01 Any thoughts on what to do with the errors? Do we want to panic?

src/backends/perf_pt/mod.rs
```diff
@@ -403,6 +403,15 @@ impl Default for PerfPTThreadTracer {
     }
 }
 
+impl Drop for PerfPTThreadTracer {
+    fn drop(&mut self) {
+        if matches!(self.state, TracerState::Started) {
```
vext01, line 408: I guess that can't be `==` or you'd have used it :P

ptersilie: I haven't tried, but last time I used `==` on an enum it didn't work.

vext01:
>  Do we want to panic?

Shall we panic for now, and we can revise that later if we think it's safe to do so?

src/backends/perf_pt/mod.rs
ptersilie: Hm, seems to work. So `matches!` is probably only needed for more complex matches.

ptersilie:
Okay, I simplified the condition and will panic upon an error. Needs squashing.

src/backends/perf_pt/mod.rs
```diff
@@ -403,6 +403,17 @@ impl Default for PerfPTThreadTracer {
     }
 }
 
+impl Drop for PerfPTThreadTracer {
+    fn drop(&mut self) {
+        if self.state == TracerState::Started {
+            // If we haven't stopped the tracer already, stop it now.
+            if let Some(err) = self.stop_tracing().err() {
```
ltratt, line 410: Is this `self.stop_tracing().unwrap()` in disguise?

ptersilie: Hah, yeah, I think it is. Why write one line when you can write three?

ltratt: OK, let's simplify it then :)

src/backends/perf_pt/mod.rs
ptersilie: Ah, that was it!

src/backends/perf_pt/mod.rs
ptersilie: Fixed in 833fba8.

ptersilie:
Ok to squash?

vext01:
Please squash

ptersilie:
Squashed!

PR #96, opened 2022-05-30
Various minor tidy ups

ltratt:
A slightly rag-bag set of commits that aim to make the hwtracer crate a bit easier to work with: separating code out into separate files; and removing what, in retrospect, is unnecessary complexity. This PR manages to delete more code than it adds!

Note I have tested this PR with `yk` and everything still builds and runs fine with these changes.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -4,14 +4,16 @@ type BlockAddr = u64;
 /// Information about a basic block.
 #[derive(Debug, Eq, PartialEq)]
 pub struct Block {
-    /// Virtual address of the first instruction in this block.
+    /// Virtual address of the start of the first instruction in this block.
     first_instr: BlockAddr,
-    /// Virtual address of the last instruction in this block.
+    /// Virtual address of the start of the last instruction in this block.
     last_instr: BlockAddr,
 }
 
 impl Block {
-    /// Creates a new basic block from a start address and a length in bytes.
+    /// Creates a new basic block from the virtual addresses of:
+    ///   * the start of the first instruction in the basic block.
+    ///   * the start of the last instruction in the basic block.
     pub fn new(first_instr: BlockAddr, last_instr: BlockAddr) -> Self {
         Self {
             first_instr,
@@ -19,12 +21,12 @@ impl Block {
         }
     }
 
-    /// Returns the virtual address of the first instruction in this block.
+    /// Returns the virtual address of the start of the first instruction in this block.
     pub fn first_instr(&self) -> BlockAddr {
         self.first_instr
     }
 
-    /// Returns the virtual address of the last instruction in this block.
+    /// Returns the virtual address of the start of the last instruction in this block.
     pub fn last_instr(&self) -> BlockAddr {
         self.last_instr
     }
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,0 +1,28 @@
+/// Information about a basic block.
+#[derive(Debug, Eq, PartialEq)]
+pub struct Block {
+    /// Virtual address of the first instruction in this block.
+    first_instr: u64,
+    /// Virtual address of the last instruction in this block.
+    last_instr: u64,
+}
+
+impl Block {
+    /// Creates a new basic block from a start address and a length in bytes.
+    pub fn new(first_instr: u64, last_instr: u64) -> Self {
+        Self {
+            first_instr,
+            last_instr,
+        }
+    }
+
+    /// Returns the virtual address of the first instruction in this block.
+    pub fn first_instr(&self) -> u64 {
+        self.first_instr
+    }
+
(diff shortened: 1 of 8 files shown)


src/block.rs
```diff
@@ -1,28 +1,31 @@
+#[cfg(target_arch = "x86_64")]
+type BlockAddr = u64;
```
vext01, line 2: The commit message for this commit isn't quite accurate.

We use a `u64` not because we only work on x86_64, but because [libipt (which only works on x86_64) uses `uint64_t` to express block addresses](https://github.com/intel/libipt/blob/c848a85c3104e2f5780741f85de5c9e65476ece2/libipt/include/intel-pt.h.in#L2381-L2389).

So this commit is really about supporting different decoders, not different architectures.

Would it make more sense to use `usize`, as in most chips, the instruction pointer is exactly wide enough to store a pointer?

vext01:
Just a handful of small comments from me.

src/block.rs
vext01: (and i'm assuming that we don't want to cater for decoding a trace for a different platform)

ltratt: A question which I haven't fully worked out the answer to is: how generic is `hwtracer` supposed to be? At the moment it supports various backends, of which libipt/PT is but one, so having libipt's notion of address being the (possibly generic) `Block` idea of address seems odd. But an alternative thought is that this library should actually be called something like "pttracer", at which point that seems less of a concern.

src/block.rs
```diff
@@ -4,27 +4,29 @@ type BlockAddr = u64;
 /// Information about a basic block.
 #[derive(Debug, Eq, PartialEq)]
 pub struct Block {
-    /// Virtual address of the first instruction in this block.
+    /// Virtual address of the start of the first instruction in this block.
```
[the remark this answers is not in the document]
ltratt: It wasn't obvious to me if it pointed to the start or end of the last instruction.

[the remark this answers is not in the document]
vext01: Hrm. But why would the address of an instruction be defined as "it's last byte"? That'd be super odd.

src/block.rs
vext01: I see what you mean. I only implemented stuff at the block-level, as that's the only granularity we care about right now.

(libipt also has a decoder that works at instruction-level granularity)

src/block.rs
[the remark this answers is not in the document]
ltratt: This is basically a contiguous vector of things, right? There are two common ways of encoding this: start address & length; first byte & last byte. What we've got with `Block` at the moment is the far less common "start address & start address of last element". Because it's unusual I think it needs documenting.

src/block.rs
ltratt: I think leaving it generic-ish for the time being is OK: that means we can later easily make it a bit more or less generic as we think is best.

src/block.rs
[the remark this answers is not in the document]
vext01: I'm not gonna die on this hill, so I'll allow it :P

ltratt:
No squashing needed!

vext01:
bors ping

vext01:
bors ping

ltratt:
`stderr: /bin/sh: 2: git: not found` huh?

vext01:
It's likely this is the first time that this repo has seen the new dockerized CI.

The [default image](https://github.com/softdevteam/buildbot_config/blob/ce71e8f0322ec7f6e6c102adb91bfaa16bce64d6/bin/run_docker_ci_job#L56) doesn't install git, so you can either add a dockerfile that does install git ([like this](https://github.com/ykjit/ykllvm/blob/3b7789ed26a209d8ec06ddb503da2465789139a0/.buildbot_dockerfile_debian#L5), or you can update the default image to include git.

Your call.

vext01:
The [build succeeded](http://ci.soft-dev.org/#/builders/8/builds/321) and github was notified, yet bors is still waiting.

The reason, `bors.toml` is waiting for the wrong builder.

Change:
```
status = [
    "buildbot/buildbot-build-script"
]
```

to:
```
status = ["buildbot/yk-buildbot-build-script-hw"]
```

ltratt:
Fixed in https://github.com/ykjit/hwtracer/pull/96/commits/95dd1c3992d93c995585c6b112669a5c8f4ee3ac.

vext01:
Bors allowed 3 minutes, but should have allowed 10:
https://github.com/ykjit/hwtracer/blob/2be4af3be5253b21d186d59675574343e0c4e496/bors.toml#L7

I'm going to retry it.

bors r+

vext01:
OK, bors is wild today.... No idea why it said "timed out" and then merged it...

ltratt:
If it works, it works!

PR #99, opened 2022-10-14
Separate trace collection from trace decoding.

vext01:
We are about to add our own trace decoder (as an alternative to libipt) to hwtracer. In order to do this we need the ability to have separate notions of trace collection and trace decoding.

In other words, instead of having a `PerfPT` backend, we now have a `PerfTraceCollector` and a `LibIPTTraceDecoder`.

(There was also something called a `Dummy` backend, which I saw no utility to, and have thus removed)

This is a large change, that will be hard to review I'm afraid.

In doing this work I've realised how crusty and unidiomatic some of the hwtracer code is. Regardless I've tried hard to resist the urge to uncrustify unrelated parts of code. I will fix some of this stuff in follow up PRs.

`yk` companion: https://github.com/ykjit/yk/pull/580


src/collect/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+/// The interface offered by all trace collectors.
+pub trait TraceCollector: Send + Sync {
+    /// Return a `ThreadTraceCollector` for tracing the current thread.
+    fn thread_collector(&self) -> Box<dyn ThreadTraceCollector>;
+}
+
+pub trait ThreadTraceCollector {
+    /// Start recording a trace.
+    ///
+    /// Tracing continues until [stop_collector](trait.ThreadTraceCollector.html#method.stop_collector) is called.
```
ltratt, line 27: I *think* you can just say "Tracing continues until [stop_collector] is called." and rustdoc will do the right thing. I might be wrong!

src/collect/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+
+impl TraceCollectorKind {
+    // Finds a suitable `TraceCollectorKind` for the current hardware/OS.
+    fn default_for_platform() -> Result<Self, ()> {
+        for kind in TraceCollectorKind::iter() {
+            if Self::match_platform(&kind).is_ok() {
+                return Ok(kind);
+            }
+        }
+        Err(()) // No suitable collector found for this system.
```
ltratt, line 50: Redundant comment.

src/collect/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Kinds of collector that hwtracer supports (in order of "auto-selection preference").
+#[derive(Debug, EnumIter)]
+pub enum TraceCollectorKind {
+    /// The `perf` subsystem, as found on Linux.
+    Perf,
+}
+
+impl TraceCollectorKind {
+    // Finds a suitable `TraceCollectorKind` for the current hardware/OS.
+    fn default_for_platform() -> Result<Self, ()> {
```
ltratt, line 44: I think this should be `Option<Self>` as there's no error implied if one isn't found.

src/collect/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+        for kind in TraceCollectorKind::iter() {
+            if Self::match_platform(&kind).is_ok() {
+                return Ok(kind);
+            }
+        }
+        Err(()) // No suitable collector found for this system.
+    }
+
+    /// Returns `Ok` if the this collector is appropriate for the current platform.
+    fn match_platform(&self) -> Result<(), HWTracerError> {
```
ltratt, line 54: Probably also `Option`.

src/collect/perf/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+            return Err(HWTracerError::BadConfig(String::from(
+                "aux_bufsize must be a positive power of 2",
+            )));
+        }
+
+        Self::check_perf_perms()?;
+        Ok(Self { config })
+    }
+
+    fn check_perf_perms() -> Result<(), HWTracerError> {
```
ltratt, line 59: This could do with a doc string and, probably, a different name because it feels odd that `check` returns `Err`.

But, looking at it, I wonder why we don't just bundle all of this code into `new`?

src/decode/libipt/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+    ) -> bool;
+    fn hwt_ipt_free_block_decoder(decoder: *mut c_void);
+    // util.c
+    pub(crate) fn hwt_ipt_is_overflow_err(err: c_int) -> bool;
+    // libipt
+    pub(crate) fn pt_errstr(error_code: c_int) -> *const c_char;
+}
+
+pub(crate) struct LibIPTTraceDecoder {
+    //trace: &'t dyn Trace,
```
ltratt, line 34: Commented code.

src/decode/libipt/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+impl TraceDecoder for LibIPTTraceDecoder {
+    fn new() -> Self {
+        Self {}
+    }
+
+    fn iter_blocks<'t>(
```
ltratt, line 42: I claim that this can be `fn iter_blocks(&self, &dyn Trace) -> Box<dyn Iterator <Item = Result<Block, HWTracerError>>>`. I might be wrong!

src/decode/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+#[cfg(decoder_libipt)]
+use libipt::LibIPTTraceDecoder;
+
+#[derive(Clone, Copy, Debug, EnumIter)]
+pub enum TraceDecoderKind {
+    LibIPT,
+}
+
+impl TraceDecoderKind {
+    fn default_for_platform() -> Result<Self, ()> {
```
ltratt, line 18: Same comments as the other `default_for_platform` function.

ltratt:
I've tried to comment on things that I think are either new-ish or are easily changed. Nothing major.

src/collect/mod.rs
vext01: The entire crate's docstrings are in dire need of attention. I was going to do a follow up PR.

ltratt: OK fair enough.

src/collect/mod.rs
vext01: I think it's so that `?` can propagate the error easily.

src/collect/mod.rs
vext01: This is also existing code that has moved. Let's address it in a follow-up PR.

src/collect/mod.rs
ltratt: `?` can work with `Option` in some cases. `Result<_, ()>` isn't very idiomatic overall.

src/collect/perf/mod.rs
ltratt: I will forget. So we can file an issue or do it as a commit in this PR. I think a commit in this PR is easiest, but it's up to you.

vext01: Good point actually. OK let's nip these in the bud now.

ltratt: Works for me!

src/collect/mod.rs
vext01: Not sure about this one. The error conveys information in the failure that `None` couldn't.

ltratt: Yes you're right.

src/collect/perf/mod.rs
vext01: Do you mind leaving this in a function? In the future we only need this check if we are going to collect a PT trace (and not e.g. some other kind of hardware trace). wrapping it around `if (PT) {` will make it a bit too-far indented for my taste.

ltratt: It's worse as it is because now *consumers* of this struct have to know if they're using PT or not! So I really think this should be merged.

src/decode/libipt/mod.rs
vext01: Just following compiler's orders guv'ner:

```
$ cargo check
    Checking hwtracer v0.1.0 (/home/vext01/research/hwtracer)
error: lifetime may not live long enough
  --> src/decode/libipt/mod.rs:51:9
   |
40 |     fn iter_blocks<'t>(
   |                    -- lifetime `'t` defined here
...
51 |         Box::new(itr)
   |         ^^^^^^^^^^^^^ returning this value requires that `'t` must outlive `'static`
   |
help: to declare that the trait object captures data from argument `self`, you can add an explicit `'t` lifetime bound
   |
43 |     ) -> Box<dyn Iterator<Item = Result<Block, HWTracerError>> + 't> {
   |                                                                ++++
```

ltratt: So maybe `'t` isn't needed but `'_` on the `Result` is?

src/collect/perf/mod.rs
vext01: See #100 -- I plan to fix this soon.

ltratt: That feels like a separate issue to me. If there's a substantive argument against merging I'm all for it but "too much indentation" isn't very persuasive ;)

src/collect/mod.rs
vext01: Fixed in [ca63e70](https://github.com/ykjit/hwtracer/pull/99/commits/ca63e70b16b138d4cbeeec9424bcc4128f4d310a)

src/collect/mod.rs
vext01: Fixed in [ca63e70](https://github.com/ykjit/hwtracer/pull/99/commits/ca63e70b16b138d4cbeeec9424bcc4128f4d310a)

src/decode/libipt/mod.rs
vext01: Fixed in [94cb536](https://github.com/ykjit/hwtracer/pull/99/commits/94cb536e85e8dbeb63fd23eae0d32389b5cbcf04)

src/decode/libipt/mod.rs
vext01: Like this?

```
-> Box<dyn Iterator<Item = Result<Block, HWTracerError>> + '_>
```

That's what we have currently.

ltratt: Without the `<'t>` and other `'t` stuff in the signature. [Maybe it's necessary but I think it might not be.]

ltratt: OK, sorry for the wild goose chase! I should learn not to be too clever with the borrow checker :)

ltratt:
Just one comment left!

src/decode/libipt/mod.rs
vext01: I tried the very same things myself when preparing this PR. The joys of Rust...

src/collect/perf/mod.rs
vext01: Fixed in 19872e6

ltratt:
Please squash.

vext01:
splat

vext01:
Just had a quick look at this.

We are maxing out the trace buffer, which is currently set (by default) to 1024 *pages* (a page is 4096 bytes by default on a typical linux system, so that's about 4MiB, by my calculations).

The test in question does in indeed test a "long" trace: it calls `SystemTime::now().elapsed().unwrap().subsec_nanos()` in a loop 3000 times.

Shall we try bumping the default trace buffer size?

vext01:
Something like that? (untested)

ltratt:
Actually, are we sure pages are a fixed size? I don't think that's true anymore, at least on Linux. We probably have to read the page size value from somewhere to calculate this.

vext01:
Page size is configurable. That's what was implied by "typical Linux system" in the commit message.

We can use the (surprisingly well-named) `getpagesize()` if you like.

ltratt:
>  We can use the (surprisingly well-named) getpagesize() if you like.

Yes we should. There's no point not doing so as we'll simply forget and confuse our future selves!

vext01:
[a0ba640](https://github.com/ykjit/hwtracer/pull/99/commits/a0ba640d797c6d93f7a8a0c4459b382e0d597e40) takes into account variable page sizes.

[978aa7f](https://github.com/ykjit/hwtracer/pull/99/commits/978aa7f88d381de1e283c5fcaae451457f7db647) fixes the intermittent test failures we spoke about in person (the commit message addresses some of the questions we had).

In doing the latter I've found a nasty bug, which I will raise an issue to explain. We should deal with this in a separate PR.

vext01:
Also pushed [af4f168](https://github.com/ykjit/hwtracer/pull/99/commits/af4f168e5d89396e14a006ae1b15a4575c44dc91) as a result of in-person conversation.

src/collect/mod.rs
```diff
@@ -25,6 +25,10 @@ const PERF_DFLT_INITIAL_TRACE_BUFSIZE: size_t = 1024 * 1024; // 1MiB
 /// The interface offered by all trace collectors.
 pub trait TraceCollector: Send + Sync {
     /// Return a `ThreadTraceCollector` for tracing the current thread.
+    ///
+    /// Note that an instance of `ThreadTraceCollector` is not thread local, and an attempt to
```
ltratt, line 29: I find this really difficult to parse. Can we simplify it to "You must not start two `TraceCollector`s in the same thread. Doing so is is undefined behaviour". Also I think this function (or the trait?) should be `unsafe`.

src/collect/perf/collect.c
```diff
@@ -66,7 +66,7 @@
 #define SYSFS_PT_TYPE   "/sys/bus/event_source/devices/intel_pt/type"
 #define MAX_PT_TYPE_STR 8
 
-#define MAX_OPEN_PERF_TRIES  20000
+#define MAX_OPEN_PERF_TRIES  50000
```
ltratt, line 69: I would suggest that adding a `pthread_yield` or `sched_yield` (or even just "sleep 100ms" or similar!) wherever this loop is will be more robust: it will make it much more likely that the other thread acquiring the lock will succeed before we try again.

vext01: We can do, although I thought yielding was kind of frowned upon? (don't ask me why, I have a vague recollection of reading something on a mailing list)

vext01: (btw, there's already a sleep: https://github.com/ykjit/hwtracer/blob/978aa7f88d381de1e283c5fcaae451457f7db647/src/collect/perf/collect.c#L70)

src/collect/mod.rs
vext01: I was thinking yesterday: why does a user facing `ThreadTraceCollector` even exist? If we had `TraceCollector::start_thread_collector()`, then this issue (I think) disappears?  There can we a hidden thread-local `ThreadTraceCollector` and the problem described in #101 I think disappear too.

src/collect/perf/collect.c
ltratt: Yielding doesn't always do what people expect (e.g. in spinlocks).

`1000 * 30` nanoseconds is 1/33333th of a second -- under such circumstances `sleep` might not even do anything! I'm fine with sleeping, but let's make it least 1/100th of a second.

src/collect/mod.rs
ltratt: I don't know the internal API, so I'm happy with whatever change you suggest (but I do want the comments to be understandable by those of us whose English is not as good as yours :)).

src/collect/perf/collect.c
vext01: Fine by me. I don't remember where these numbers even came from TBH.

src/collect/mod.rs
vext01: I'm actually OK with making a slightly stronger restriction than what the (current) API allows, so how about something really simple, like:

> You may only call `thread_tracer()` once per thread. Calling it more than once per-thread will result in undefined behavior.

This will force the user to only ever have one hanging around.

ltratt: > You may only call thread_tracer() once per thread.

Does that mean that you can only trace one *ever* in the lifetime of a thread?

vext01: Not sure what you mean. One what?

ltratt: Oops, "once ever".

In other words, can I start/stop a trace (thread_tracer?) and then, once it's stopped, start it up again? Or must I create a new thread_tracer?

vext01: The comment above is forcing the user to re-use the same thread tracer and never create a second for subsequent tracing sessions (on that thread).

ltratt: That's not how I read it. To me it says "you can only use this thread_tracer once, ever. If you want to trace again, create another thread_tracer."

vext01: The method `thread_collector()` (sorry, I used its old name `thread_tracer()` above) only obtains a thread collector. It does not tracing until asked to with `start_collector()`.

The comment is saying -- you may only obtain a thread collector once (but the collector is reusable).

ltratt: > you may only obtain a thread collector once (but the collector is reusable).

Let's say that :)

src/collect/perf/collect.c
ltratt: Personally I'd use at least 1/100th of a second, perhaps as much as 1/10th. In the grand scheme of things, small delays like that don't really slow down the test suite to any noticeable degree.

src/collect/mod.rs
vext01: I've just realised that the constraint we are documented is already violated by the test suite --  `cargo test` will use threads, and individual tests will be spinning up new instances of `ThreadTraceCollector` willy-nilly.

I think we have to roll with something like:

> A thread may obtain multiple `ThreadTraceCollector`s but must only collect a trace with one at a time.

Is that any good?

[this API is a mess -- I wish i had designed it better when we started]

ltratt: Hang on, are we changing the documentation to reflect what we *want* to happen vs. changing the documentation to reflect what *will* happen?

vext01: > you may only obtain a thread collector once (but the collector is reusable).

overly restricts the user at the cost of being very easy to explain. In other words, there are cases where it is safe to obtain > 1 thread collector, but if we tell the user they must obtain only one, it rules all unsafe cases out.

The test suite cannot abide by these rules though.

> A thread may obtain multiple ThreadTraceCollectors but must only collect a trace with one at a time.

precisely describes when it is safe to obtain >1 thread collector.

I hope that makes sense?

vext01: [The more I see the confusion, the more I want to fix the API]

ltratt: If we can fix the API without difficulty, that's probably more productive than trying to work out how to encode a complex condition in prose!

vext01: I don't think it will take that long to fix, but I would like to do it as a fresh PR if that's OK.

Are you OK with putting a placeholder comment here for now if I promise that my next PR fixes the API?

We could use:

> A thread may obtain multiple ThreadTraceCollectors but must only collect a trace with one at a time.

Even if it's a little weak.

vext01: (Actually I already have a boring docstring branch to go in after this one -- if that one could go in next, I could save time resolving conflicts later)

vext01:
I have a branch with a preliminary API redesign now. Let's move this one forward?

ltratt:
Please squash.

vext01:
splat.

vext01:
```
  ┌─ unicode-ident 1.0.5 (registry+https://github.com/rust-lang/crates.io-index):4:13
  │
4 │ license = "(MIT OR Apache-2.0) AND Unicode-DFS-2016"
  │            -^^^----^^^^^^^^^^------^^^^^^^^^^^^^^^^
  │            ││      │               │
  │            ││      │               rejected: not explicitly allowed
  │            ││      accepted: license is explicitly allowed
  │            │accepted: license is explicitly allowed
  │            license expression retrieved via Cargo.toml `license`
  │
```

vext01:
OK, so strum has pulled in this unicode license and the ISC license.

with the [unicode one](https://spdx.org/licenses/Unicode-DFS-2016.html) says:
```
 either

    (a) this copyright and permission notice appear with all copies of the Data Files or Software, or
    (b) this copyright and permission notice appear in associated Documentation.
```

Are we a "copy of the software"?

ltratt:
IMHO this license is incredibly unclear (I've encountered it in another project) but the only plausible thing we can do is to allow it in `cargo deny` and hope that we've done the right thing.

vext01:
Or eliminate strum from our code. Which would be a shame...

vext01:
OK with those fixes?

ltratt:
Please squash.

vext01:
splat.

PR #103, opened 2022-10-21
Make the API more thread-safe.

### LINKED ISSUE #101: `ThreadTraceCollector`s are not thread local.
The current (after #99 is merged) design for collecting a trace involves a `TraceCollector` and a `ThreadTraceCollector`. When you call `thread_collector()` on a `TraceCollector` you get a `ThreadTraceCollector`.

`TraceCollector` is `Send` and `Sync`, whereas `ThreadTraceCollector` is intended to be unique to a thread.

I've notice that if you call `thread_collector()` repeatedly from the same thread, you will always get a fresh instance of a `ThreadTraceCollector` each time and each fresh instance is assumed to be capable of starting an independent trace collection "session". This of course implies that you can trace any given thread more than once concurrently. You can't! Perf can't do that!

Let me demonstrate with code:

```
let thr1 = tc.thread_collector();
let thr2 = tc.thread_collector();

thr1.start_collector().unwrap(); // succeeds
thr2.start_collector().unwrap(); // also succeeds, despite the current thread already being traced!
```

(There's also some odd interaction when you come to stop the separate thread collectors, which I don't yet understand)

An obvious attempt to fix this would be to make the thread collector a thread local, and have subsequent calls to `thread_collector()` from the same thread return the same thread local collector.

The problem with this approach is:

```
let tc1: TraceCollector = ...; // collector with one configuration.
let tc2: TraceCollector = ...; // collector with a different configuration.
let thr1 = tc1.thread_collector();
let thr2 = tc2.thread_collector();
```

By the proposed fix, `thr2` is an alias for `thr1` and `tc2` gave us a thread collector with an unexpected configuration!

vext01:
Before this change, `ThreadTraceCollector` was exposed to the consumer of the hwtracer API. Calling `TraceCollector::thread_tracer()` would give back a fresh instance each time. This API would allow the user request multiple `ThreadTraceCollector`s on the same thread and start them all "collecting traces" concurrently. This leads to chaos as any given thread can only be traced at most once at a time.

Until now we've swept this under the rug, warning the user not to do the naughty thing unless they want to invoke UB, but the API still allows it to happen.

This change hides `ThreadTraceCollector`s from the public API entirely, leaving `TraceCollector`s to manage thread local instances of `ThreadTraceCollector`.

The user now starts and stops trace collection of a thread with `TraceCollector::start_thread_collector()` and
`TraceCollector::stop_thread_collector()`. `ThreadTraceCollector`s are still there behind the scenes, but the user no longer sees them and thus can't start two tracing the same thread any more.

One subtle side-effect of this new API is that we can no longer automatically stop a running trace collector automatically when the `ThreadTraceCollector` goes out of scope (via `Drop`). Since `ThreadTraceCollector`s are stored (and owned) by thread locals, they never really "fall out of scope", so the user must ensure they consistently stop trace collection before trying to re-trace the same thread.

It may be tempting to stop the current thread's `ThreadTraceCollector` when a `TraceCollector` goes out of scope, but then we'd have to track which `TraceCollector` started the `ThreadTraceCollector` and only stop the collector if they match. This starts to sound pretty over the top, and I think the proposed API is already a lot better than what we had before. Further `TraceCollector::start_thread_collector()` will return a meaningful error message if the user has forgotten to stop a collector, so it shouldn't be hard to track down instances of this blunder.

Fixes #101.

### DIFF
--- a/src/collect/mod.rs
+++ b/src/collect/mod.rs
@@ -3,7 +3,7 @@
 use crate::{errors::HWTracerError, Trace};
 use core::arch::x86_64::__cpuid_count;
 use libc::{size_t, sysconf, _SC_PAGESIZE};
-use std::{convert::TryFrom, sync::LazyLock};
+use std::{cell::RefCell, convert::TryFrom, sync::LazyLock};
 use strum::IntoEnumIterator;
 use strum_macros::EnumIter;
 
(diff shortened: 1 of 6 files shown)


vext01:
Companion PR coming soon.

src/collect/mod.rs
```diff
@@ -22,19 +22,59 @@ static PERF_DFLT_AUX_BUFSIZE: LazyLock<size_t> = LazyLock::new(|| {
...
-    ///
-    /// A thread may obtain multiple `ThreadTraceCollector`s but must only collect a trace with one
-    /// at a time.
-    ///
-    /// FIXME: This API needs to be fixed:
-    /// https://github.com/ykjit/hwtracer/issues/101
+thread_local! {
+    /// When `Some` holds the `ThreadTraceCollector` that is collecting a trace of the current
+    /// thread.
+    static THREAD_TRACE_COLLECTOR: RefCell<Option<Box<dyn ThreadTraceCollector>>> = RefCell::new(None);
```
vext01, line 28: By the way, I think I can make this `RefCell` a `Cell` later.

ltratt: That's unlikely because `Cell` requires its contents to be `Copy`.

src/collect/mod.rs
```diff
@@ -22,19 +22,59 @@ static PERF_DFLT_AUX_BUFSIZE: LazyLock<size_t> = LazyLock::new(|| {
...
+}
+
+/// The private innards of a `TraceCollector`.
+pub(crate) trait TraceCollectorImpl: Send + Sync {
     unsafe fn thread_collector(&self) -> Box<dyn ThreadTraceCollector>;
 }
 
-pub trait ThreadTraceCollector {
+/// The public interface offered by all trace collectors.
+pub struct TraceCollector {
```
ltratt, line 37: It's OK if we do `Box` but if we expect `col_impl` to live longer than `TraceCollector` it might be worth doing:

```
struct TraceCollector<'a> {
   &'a dyn TraceCollectorImpl
}
```

I don't feels strongly about this though.

src/collect/mod.rs
vext01: Ah, maybe not then.

src/collect/mod.rs
vext01: Since it's the responsibility of the `TraceCollector` to both create and and drop `ThreadTraceCollector`s, I think an owned value makes more sense.

src/collect/mod.rs
ltratt: I'll mark this one resolved :)

src/collect/mod.rs
vext01: Hold on, that's nonsense. My mistake.

If we were to store a reference to the `TraceCollectorImpl`, where would the owned value be stored? `TraceCollectorImpl` is deliberately private to the API consumer, so we would probably end up jumping through hoops to make such a design work.

ltratt: From a privacy perspective, `Box<TraceCollectorImpl>` and `&'a TraceCollectorImpl` are equivalent. Maybe I'm not understanding something?

vext01: If you store a reference, then the owned value has to live somewhere else. My question is, where would this be?

ltratt: "if we expect col_impl to live longer than TraceCollector". Do we? It's your code, not mine :)

vext01: `TraceCollector` and `TraceCollectorImpl` are part of the same conceptual object -- they are only separated to create a barrier between what's public and what's private. I expect them to be created together and to die together.

ltratt: Then what's the problem with `&'a`?

vext01: Then the owned (private) inner would have to live outside of `TraceCollector` somewhere else. Probably the `TraceCollectorBuilder` would have to own it and give out a reference to it. This would tie the lifetime of the `TraceCollector` to the `TraceCollectorBuilder`. That would be a weird API.

vext01:
No squashing needed, but did a rebase and fixed typo in commit message.

PR #105, opened 2022-10-31
Start work on a PT packet decoder.

vext01:
First commit is the meat of the work.

Other two are unrelated fixes I found along the way.

### DIFF
--- a/src/decode/mod.rs
+++ b/src/decode/mod.rs
@@ -9,9 +9,15 @@ pub(crate) mod libipt;
 #[cfg(decoder_libipt)]
 use libipt::LibIPTTraceDecoder;
 
+#[cfg(decoder_ykpt)]
+mod ykpt;
+#[cfg(decoder_ykpt)]
+use ykpt::YkPTTraceDecoder;
+
 #[derive(Clone, Copy, Debug, EnumIter)]
 pub enum TraceDecoderKind {
     LibIPT,
+    YkPT,
 }
 
 impl TraceDecoderKind {
@@ -29,10 +35,16 @@ impl TraceDecoderKind {
     fn match_platform(&self) -> Result<(), HWTracerError> {
         match self {
             Self::LibIPT => {
-                #[cfg(not(decoder_libipt))]
-                return Err(HWTracerError::DecoderUnavailable(Self::LibIPT));
                 #[cfg(decoder_libipt)]
                 return Ok(());
+                #[cfg(not(decoder_libipt))]
+                return Err(HWTracerError::DecoderUnavailable(Self::LibIPT));
+            }
+            Self::YkPT => {
+                #[cfg(decoder_ykpt)]
+                return Ok(());
+                #[cfg(not(decoder_ykpt))]
+                return Err(HWTracerError::DecoderUnavailable(Self::YkPT));
             }
         }
     }
@@ -76,7 +88,18 @@ impl TraceDecoderBuilder {
     pub fn build(self) -> Result<Box<dyn TraceDecoder>, HWTracerError> {
         self.kind.match_platform()?;
         match self.kind {
-            TraceDecoderKind::LibIPT => Ok(Box::new(LibIPTTraceDecoder::new())),
+            TraceDecoderKind::LibIPT => {
+                #[cfg(decoder_libipt)]
+                return Ok(Box::new(LibIPTTraceDecoder::new()));
+                #[cfg(not(decoder_libipt))]
+                return Err(HWTracerError::DecoderUnavailable(self.kind));
+            }
+            TraceDecoderKind::YkPT => {
+                #[cfg(decoder_ykpt)]
+                return Ok(Box::new(YkPTTraceDecoder::new()));
+                #[cfg(not(decoder_ykpt))]
+                return Err(HWTracerError::DecoderUnavailable(self.kind));
+            }
         }
     }
 }
--- a/src/decode/ykpt/mod.rs
+++ b/src/decode/ykpt/mod.rs
@@ -1,0 +1,61 @@
+//! The Yk PT trace decoder.
+
+use crate::{decode::TraceDecoder, errors::HWTracerError, Block, Trace};
+
+mod packet_parser;
+use packet_parser::PacketParser;
+
+pub(crate) struct YkPTTraceDecoder {}
+
+impl TraceDecoder for YkPTTraceDecoder {
+    fn new() -> Self {
+        Self {}
+    }
+
+    fn iter_blocks<'t>(
+        &'t self,
+        trace: &'t dyn Trace,
+    ) -> Box<dyn Iterator<Item = Result<Block, HWTracerError>> + '_> {
+        let itr = YkPTBlockIterator {
+            errored: false,
+            parser: PacketParser::new(trace.bytes()),
+        };
+        Box::new(itr)
+    }
+}
+
+/// Iterate over the blocks of an Intel PT trace using the fast Yk PT decoder.
+struct YkPTBlockIterator<'t> {
+    /// Set to true when an error has occured.
+    errored: bool,
+    /// PT packet iterator.
+    parser: PacketParser<'t>,
+}
+
+impl<'t> Iterator for YkPTBlockIterator<'t> {
+    type Item = Result<Block, HWTracerError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if !self.errored {
+            // FIXME: For now this is dummy code to prevent dead-code warnings. Later block binding
+            // logic will go here.
+            self.parser.next().unwrap().unwrap();
+        }
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        collect::TraceCollectorBuilder,
+        decode::{test_helpers, TraceDecoderKind},
+    };
+
+    #[ignore] // FIXME
+    #[test]
+    fn ten_times_as_many_blocks() {
+        let tc = TraceCollectorBuilder::new().build().unwrap();
+        test_helpers::ten_times_as_many_blocks(tc, TraceDecoderKind::YkPT);
+    }
+}
--- a/src/decode/ykpt/packet_parser/mod.rs
+++ b/src/decode/ykpt/packet_parser/mod.rs
@@ -1,0 +1,295 @@
+//! A packet parser for the Yk PT trace decoder.
+
+use crate::errors::HWTracerError;
+use deku::{bitvec::BitSlice, DekuRead};
+use std::iter::Iterator;
+
+mod packets;
+use packets::*;
+
+#[derive(Clone, Copy, Debug)]
+enum PacketParserState {
+    /// Initial state, waiting for a PSB packet.
+    Init,
+    /// The "normal" decoding state.
+    Normal,
+    /// We are decoding a PSB+ sequence.
+    PSBPlus,
+}
+
+impl PacketParserState {
+    /// Returns the kinds of packet that are valid for the state.
+    fn valid_packets(&self) -> &'static [PacketKind] {
+        // Note that the parser will attempt to match packet kinds in the order that they appear in
+        // the returned slice. For best performance, the returned slice should be sorted, most
+        // frequently expected packet kinds first.
+        //
+        // OPT: The order below is a rough guess based on what limited traces I've seen. Benchmark
+        // and optimise.
+        match self {
+            Self::Init => &[PacketKind::PSB],
+            Self::Normal => &[
+                PacketKind::ShortTNT,
+                PacketKind::PAD,
+                PacketKind::FUP,
+                PacketKind::TIP,
+                PacketKind::CYC,
+                PacketKind::LongTNT,
+                PacketKind::PSB,
+                PacketKind::MODE,
+                PacketKind::TIPPGE,
+                PacketKind::TIPPGD,
+            ],
+            Self::PSBPlus => &[PacketKind::CBR, PacketKind::PSBEND],
+        }
+    }
+
+    /// Check if the parser needs to transition to a new state as a result of parsing a certain
+    /// kind of packet.
+    fn transition(&mut self, pkt_kind: PacketKind) {
+        let new = match (*self, pkt_kind) {
+            (Self::Init, PacketKind::PSB) => Self::PSBPlus,
+            (Self::Normal, PacketKind::PSB) => Self::PSBPlus,
+            (Self::PSBPlus, PacketKind::PSBEND) => Self::Normal,
+            _ => return, // No state transition.
+        };
+        *self = new;
+    }
+}
+
+pub(super) struct PacketParser<'t> {
+    /// The raw bytes of the PT trace we are iterating over.
+    bytes: &'t [u8],
+    /// The parser operates as a state machine. This field keeps track of which state we are in.
+    state: PacketParserState,
+    /// The most recent Target IP (TIP) value that we've seen. This is needed because updated TIP
+    /// values are sometimes compressed using bits from the previous TIP value.
+    prev_tip: usize,
+}
+
+/// Attempt to read the packet of type `$packet` using deku. On success wrap the packet up into the
+/// corresponding discriminant of `Packet`.
+macro_rules! read_to_packet {
+    ($packet: ty, $bits: expr, $discr: expr) => {
+        <$packet>::read($bits, ()).and_then(|(r, p)| Ok((r, $discr(p))))
+    };
+}
+
+/// Same as `read_to_packet!`, but with extra logic for dealing with packets which encode a TIP.
+macro_rules! read_to_packet_tip {
+    ($packet: ty, $bits: expr, $discr: expr, $prev_tip: expr) => {
+        <$packet>::read($bits, ()).and_then(|(r, p)| {
+            let ret = if p.needs_prev_tip() {
+                Ok((r, $discr(p, Some($prev_tip))))
+            } else {
+                Ok((r, $discr(p, None)))
+            };
+            ret
+        })
+    };
+}
+
+impl<'t> PacketParser<'t> {
+    pub(super) fn new(bytes: &'t [u8]) -> Self {
+        Self {
+            bytes,
+            state: PacketParserState::Init,
+            prev_tip: 0,
+        }
+    }
+
+    /// Attempt to parse a packet of the specified `PacketKind`.
+    fn parse_kind(&mut self, kind: PacketKind) -> Option<Packet> {
+        let bits = BitSlice::from_slice(self.bytes).ok()?;
+        let parse_res = match kind {
+            PacketKind::PSB => {
+                read_to_packet!(PSBPacket, bits, Packet::PSB)
+            }
+            PacketKind::CBR => read_to_packet!(CBRPacket, bits, Packet::CBR),
+            PacketKind::PSBEND => read_to_packet!(PSBENDPacket, bits, Packet::PSBEND),
+            PacketKind::PAD => read_to_packet!(PADPacket, bits, Packet::PAD),
+            PacketKind::MODE => read_to_packet!(MODEPacket, bits, Packet::MODE),
+            PacketKind::TIPPGE => {
+                read_to_packet_tip!(TIPPGEPacket, bits, Packet::TIPPGE, self.prev_tip)
+            }
+            PacketKind::TIPPGD => {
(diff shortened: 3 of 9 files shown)


src/decode/mod.rs
```diff
@@ -9,9 +9,15 @@ pub(crate) mod libipt;
...
 
+#[cfg(decoder_ykpt)]
+mod ykpt;
+#[cfg(decoder_ykpt)]
+use ykpt::YkPTTraceDecoder;
+
 #[derive(Clone, Copy, Debug, EnumIter)]
 pub enum TraceDecoderKind {
     LibIPT,
+    YkPT,
```
ltratt, line 20: Shouldn't this be `#[cfg(decoder_ykpt)]` too?

src/decode/mod.rs
```diff
@@ -34,6 +40,7 @@ impl TraceDecoderKind {
                 #[cfg(decoder_libipt)]
                 return Ok(());
             }
+            Self::YkPT => return Ok(()),
```
ltratt, line 43: `return` isn't needed here.

src/decode/mod.rs
```diff
@@ -77,6 +84,12 @@ impl TraceDecoderBuilder {
         self.kind.match_platform()?;
         match self.kind {
             TraceDecoderKind::LibIPT => Ok(Box::new(LibIPTTraceDecoder::new())),
+            TraceDecoderKind::YkPT => {
+                #[cfg(decoder_ykpt)]
+                return Ok(Box::new(YkPTTraceDecoder::new()));
```
ltratt, line 94: `return` not needed

src/decode/ykpt/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+            errored: false,
+            parser: PacketParser::new(trace.bytes()),
+        };
+        Box::new(itr)
+    }
+}
+
+/// Iterate over the blocks of an Intel PT trace using the fast Yk PT decoder.
+struct YkPTBlockIterator<'t> {
+    /// Error state. Set to true when an error occurs, thus invalidating the iterator.
```
ltratt, line 29: Could just be `/// Set to true when an error has occurred`?

src/decode/ykpt/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// PT packet iterator.
+    parser: PacketParser<'t>,
+}
+
+impl<'t> Iterator for YkPTBlockIterator<'t> {
+    type Item = Result<Block, HWTracerError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // There was an error in a previous iteration.
+        if self.errored {
```
ltratt, line 40: I tend to prefer:

```rust
if !self.errored {
  // common case
} else {
  // uncommon case
}
```

on the basis that compilers *tend* to guess that the "true" branch is the common case.

vext01: The idiom was copied from the [libipt decoder](https://github.com/ykjit/hwtracer/blob/9e517be439aaad9b5b8ffeac2b1b8d5f0860c6f2/src/decode/libipt/mod.rs#L119), where it makes more sense as the method is long and you don't want so much indented code.

I expect this function will grow too, but I wouldn't worry too much right now as we don't actually cover this code path yet. It's only implemented to satisfy a trait at the moment.

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+    Init,
+    /// The "normal" decoding state.
+    Normal,
+    /// We are decoding a PSB+ sequence.
+    PSBPlus,
+}
+
+impl PacketParserState {
+    /// Returns the kinds of packet that are valid for the state.
+    fn valid_packets(&self) -> &'static [PacketKind] {
```
ltratt, line 22: I'm not convinced by this function because in general I expect the follow-on packets to be somewhat conditional and then the `&'static` lifetimes won't work. I would expect that this will need to be inlined but, perhaps, PT is simple enough to get away with this?

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+                PacketKind::PSB,
+                PacketKind::MODE,
+                PacketKind::TIPPGE,
+                PacketKind::TIPPGD,
+            ],
+            Self::PSBPlus => &[PacketKind::CBR, PacketKind::PSBEND],
+        }
+    }
+
+    /// Check if we need to transition state after parsing a given packet kind.
```
ltratt, line 47: I can't parse this.

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    fn mk_parse_error(&mut self) -> HWTracerError {
+        let mut s = String::new();
+        write!(
+            &mut s,
+            "In state '{:?}', failed to parse packet: ",
+            self.state
+        )
+        .unwrap();
```
ltratt, line 107: Isn't this function just an odd way of saying `format!("In state {:?}, failed to parse packet {}", self.state, self.byte_stream_str(...))`?

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        write!(&mut s, "{}", self.byte_stream_str(8, ", ")).unwrap();
+        HWTracerError::TraceParseError(s)
+    }
+
+    /// Attempt to parse a packet of the specified `PacketKind`.
+    fn parse_kind(&mut self, kind: PacketKind) -> Option<Packet> {
+        let bits = BitSlice::from_slice(self.bytes).ok()?;
```
ltratt, line 114: `ok` looks suspicious here, as it's normally used to mean "ignore errors".

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+            PacketKind::TIPPGD => {
+                read_to_packet_tip!(TIPPGDPacket, bits, Packet::TIPPGD, self.prev_tip)
+            }
+            PacketKind::ShortTNT => read_to_packet!(ShortTNTPacket, bits, Packet::ShortTNT),
+            PacketKind::LongTNT => read_to_packet!(LongTNTPacket, bits, Packet::LongTNT),
+            PacketKind::TIP => read_to_packet_tip!(TIPPacket, bits, Packet::TIP, self.prev_tip),
+            PacketKind::FUP => read_to_packet_tip!(FUPPacket, bits, Packet::FUP, self.prev_tip),
+            PacketKind::CYC => read_to_packet!(CYCPacket, bits, Packet::CYC),
+        };
+        if let Ok((remain, pkt)) = parse_res {
```
ltratt, line 135: This is probably shorter as a `match`.

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        for i in 0..nbytes {
+            vals.push(format!("{:08b}", self.bytes[i]));
+        }
+
+        if self.bytes.len() > nbytes {
+            vals.push("...".to_owned());
+        }
+
+        let mut ret = String::new();
+        write!(ret, "{}", vals.join(sep)).unwrap();
```
ltratt, line 173: The `write!` is pointless: `vals.join` creates a string which we then `write!` into an empty string!

src/decode/ykpt/packet_parser/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        Ok(pkt)
+    }
+}
+
+impl<'t> Iterator for PacketParser<'t> {
+    type Item = Result<Packet, HWTracerError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.bytes.is_empty() {
```
ltratt, line 198: I'd prefer the common case to be first (as above).

src/decode/ykpt/packet_parser/packets.rs
```diff
@@ -0,0 +1,355 @@
...
+    val: u8,
+}
+
+impl IPBytes {
+    #[cfg(test)]
+    pub(in crate::decode::ykpt) fn new(val: u8) -> Self {
+        debug_assert!(val >> 3 == 0);
+        Self { val }
+    }
+    /// Returns `true` if we need the previous TIP value to make sense of the new one.
```
ltratt, line 21: Missing blank line.

src/decode/ykpt/packet_parser/packets.rs
```diff
@@ -0,0 +1,355 @@
...
+    ) -> Option<usize> {
+        let res = match ip_bytes.val {
+            0b000 => {
+                debug_assert!(matches!(self, Self::OutOfContext));
+                return None;
+            }
+            0b001 => {
+                // The result is bytes 63..=16 from `prev_tip` and bytes 15..=0 from `ip`.
+                if let Self::Ip16(v) = self {
+                    prev_tip.unwrap() & 0xffffffffffff0000 | usize::try_from(*v).unwrap()
```
ltratt, line 77: Is this meant to be `usize` or `u64`?

src/decode/ykpt/mod.rs
ltratt: Updating it in both places in this PR is fine.

src/decode/ykpt/packet_parser/mod.rs
vext01: > I expect the follow-on packets to be somewhat conditional and then the &'static lifetimes won't work.

Can you elaborate. What's a follow-on packet? What's a conditional packet?

This function just says what *kinds* of packet are valid in each state, so that we don't waste effort decoding a packet which can never arise in the current parsing context. It isn't returning any packet data, so I do expect this to only ever return a static array.

src/decode/ykpt/mod.rs
ltratt: The problem with perpetuating the bad idiom is that it gives us permission to perpetuate it again (and again (and again)). It would be better to fix this now IMHO.

src/decode/ykpt/packet_parser/mod.rs
ltratt: s/follow-on/valid/

I'd be moderately surprised if for a state `S` we can always statically calculate the valid packets with 100% accuracy (but maybe I'm wrong!).

src/decode/mod.rs
vext01: But it is. It's a limitation/bug in Rust's parser.

```
   Compiling hwtracer v0.1.0 (/home/vext01/research/hwtracer)
error: expected `;`, found `#`
  --> src/decode/mod.rs:94:54
   |
94 |                 Ok(Box::new(YkPTTraceDecoder::new()))
   |                                                      ^ help: add `;` here
95 |                 #[cfg(not(decoder_ykpt))]
   |                 - unexpected token
```

ltratt: Egads!

src/decode/ykpt/packet_parser/mod.rs
vext01: Parses fine here, but we can expand it to something more explicit:

> Check if the parser needs to transition to a new state as a result of parsing a certain kind of packet.

ltratt: Your rephrasing is much better!

src/decode/ykpt/packet_parser/mod.rs
vext01: `ok` converts from `Result` to `Option`.

I've used it here because I don't see value in propagating the error from the bitvec crate up, when `None` is sufficient for our purposes.

ltratt: My point is: why would we get an `Err` and why are we ignoring it? Ignoring it without a comment explaining why raises so many questions!

src/decode/ykpt/packet_parser/mod.rs
vext01: Oops. That's a left-over from an earlier design where the output string was passed in. Wlil fix.

src/decode/ykpt/packet_parser/packets.rs
vext01: I'm sorry, did you say *missing* blank line? I feel like I don't even know you any more :rofl: 

Will fix.

ltratt: As I typed it, I realised that my fingers are not used to that pattern 😆

ltratt:
In the commit message this doesn't parse:

> This change is the result several different designs using different
methods for bitwise binary packet parsing.

And I must admit that merely adding "of" doesn't make it make sense to me. Maybe I'm missing out on something!

src/decode/ykpt/packet_parser/packets.rs
vext01: I knew you'd flag this :)

I chose `u64` for the compressed address and `usize` for the decompressed address.

Justification:
 - Before decompression the address is just a field in a packet. It wouldn't be right to call that an address, so usize makes little sense there.
 - After decompression we have a `RIP` (instruction pointer register) value, and it's also a virtual address. This is by definition a pointer-sized integer and `usize` is perfect. In theory, PT works in 32-bit mode (on a 64-bit machine) too, so there you wouldn't want a `u64`.

Do you buy it?

ltratt: I do!

src/decode/ykpt/packet_parser/mod.rs
vext01: There's no calculation required. The Intel manual specifies when a given type of packet may come in.

vext01: Just to show that I'm not making this up, here's an example:
image

src/decode/ykpt/packet_parser/mod.rs
vext01: It used to be called from many places, but I see there is now only one call site. We can inline it, yes.

src/decode/ykpt/packet_parser/mod.rs
vext01: I don't see it as ignoring it. We return the exceptional value `None` to indicate the error, but I it's not interesting to know why it failed.

FWIW:
https://docs.rs/bitvec/0.22.3/bitvec/slice/struct.BitSlice.html#returns-2

If you really want to propagate it, we can, but we'd have to either make our own composite error type (compounding BitVec and deku error types), or add those error types to `HWTracerError`.

Note that we do the same below where we discard a deku error and return None.

ltratt: Ah, I see. OK with me!

src/decode/ykpt/packet_parser/mod.rs
ltratt: OK, let's worry about it later. I don't like the current design, but that's because I have written too many parsers, and this is a parser in disguise, but it's not too important now.

src/decode/mod.rs
vext01: No, the user should still be able to request any given decoder, even if it isn't available.

src/decode/ykpt/packet_parser/mod.rs
vext01: It turns out 7 lines instead of 6:

```
        if let Ok((remain, pkt)) = parse_res {
            self.bytes = remain.as_raw_slice();
            Some(pkt)
        } else {
            None
        }
```

vs. 

```
        match parse_res {
            OK((remain, pkt)) => {
                self.bytes = remain.as_raw_slice();
                Some(pkt)
            },
            Err(_) => None,
        }
```

vext01:
> In the commit message this doesn't parse

How about something like:

> I arrived at the approach you see in this change after trying several other, less-satisfying, ways of parsing binary packets (with bitwise field granularity).

src/decode/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/mod.rs
vext01: New code fix in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

Existing code fix in [09ce652](https://github.com/ykjit/hwtracer/pull/105/commits/09ce652c772e616b5931dcebf31c359e5de2e1c6)

src/decode/ykpt/packet_parser/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/packet_parser/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/packet_parser/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/packet_parser/mod.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

src/decode/ykpt/packet_parser/packets.rs
vext01: Fixed in [3567abf](https://github.com/ykjit/hwtracer/pull/105/commits/3567abfb722b6106c105a07bc9a0b4ceda47d3a4)

vext01:
Over to you.

ltratt:
Next time, can we split fixes up into smaller commits please? It makes reviewing a *lot* easier.

src/decode/mod.rs
ltratt: What does it mean to request a decoder that isn't available?

src/decode/ykpt/packet_parser/mod.rs
ltratt: OK.

src/decode/mod.rs
vext01: The decoders are conditionally compiled in, but we still allow the user of the API to request any of them when instantiating a decoder.

So if the user does:
```
TraceDecoderBuilder::new().kind(TraceDecoderKind::LibIPT).build();
```

And `hwtracer` was built without libipt, then this would return `Err(HWTracerError::DecoderNotAvailable)`.

If we were to make (e.g.) `TraceDecoderKind::LibIPT` conditionally compiled in too, then that would force consumers of our API to also have compile time switches for hwtracer's decoders and encoders in their build systems too, thus spreading compile time knobs up the dependency chain. This seems undesirable to me.

ltratt: It's not important now but FWIW, I think trading compile-time errors for run-time errors is suboptimal: I would rather know at compile-time that my code won't do what I hope than wait until runtime.

ltratt:
Please squash.

src/decode/mod.rs
vext01: That's fine as long as you don't mind imposing conditional compilation on the consumers of the library.

If `DecoderKind::LibIPT` isn't compiled in, then that constant can't be used in consumers, and they too will have to conditionally compile it's use. I think that's a poor API personally.

If you still feel strongly, we can change it later.

ltratt: > That's fine as long as you don't mind imposing conditional compilation on the consumers of the library.

I not only don't mind, I actively *want* to do this!

But, yes, a later PR is fine for this.

vext01:
Squashed, an updated commit message.

